PR #13, opened 2018-06-18
PXD-396 Feat/graphql query

macintoshpie:
- added tests for graphql
- refactored some utility functions in sheepdogAPI


assertions/peregrineAssertions.js
```diff
@@ -0,0 +1,114 @@
...
+        ++fields_compared;
+        expect(node_data[field]).to.equal(q_res[field]);
+      }
+    } catch(e) {
+      fail_list.push(e)
+    }
+  });
+
+  expect(fields_compared).to.be.above(0, 'failed to compare node to result: all fields were different');
+  expect(fail_list).to.have.lengthOf(0);
```
thanh-nguyen-dang: should use `expect(fail_list).to.deep.equal([]);` so that the content of the list will be shown when the test fail

macintoshpie: Ah nice, didn't know about that! I'll make that change

PR #21, opened 2018-07-19
Fix/selenium start

macintoshpie:
* wait for selenium server to start before running tests


macintoshpie:
Receiving similar error to the folks over here: https://github.com/webdriverio/wdio-selenium-standalone-service/issues/23
Not sure that my changes will fix it, but I've also added some debugging text to try and see what is going on when the error is repeated.

PR #31, opened 2018-08-07
PXD-1383 Feat/core metadata

paulineribeyre:
Add landing page test

### DIFF
--- a/suites/portal/coremetadatapageTest.js
+++ b/suites/portal/coremetadatapageTest.js
@@ -1,14 +1,42 @@
 'use strict';
 
+let nodes, nodes_list;
+
 Feature('Core Metadata');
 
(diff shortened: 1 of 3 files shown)


suites/portal/coremetadatapageTest.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+Feature('Core Metadata');
+
+Scenario('test core metadata', function* (I) {
```
macintoshpie: This is a good test! I like how you abstracted away navigation to the page. But we need a few more Scenarios probably to cover all of the cases. Are there special cases, ie the file was not found, or I don't have permission to access it, or there is no data in the project? Also I wonder if they wanted you to test the filters on the side, not sure, probably should ask about that.

PR #39, opened 2018-08-20
refactor(readme): docs for setup and writing tests

macintoshpie:


README.md
```diff
@@ -1,8 +1,160 @@
...
+```
+npm test -- --grep "@MyTag"
+```
+You can also run a specific suite by doing the following:
+```
+npm test ./suites/apis/myApiTest.js
+```
+
+## Writing Tests
+Tests are single .js files that test a group of closely related features. They are stored in `suites/apis` and `suites/portal`, with filenames matching the pattern `*Test.js`.
```
thanh-nguyen-dang: Could you add the more high-level explanation about TestSuite (what it is, how it should look like, what it consists of). Then explain a bit more about the set of question, task, sequence, ... when you write a test suite.

PR #41, opened 2018-09-17
Refactor/responses

macintoshpie:
use Gen3Response wrapper for all api responses


themarcelor:
Ancient PR. Abandoning it.

PR #56, opened 2018-11-14
tests(google): add missing google sa tests and clarify scenarios [PXD-2277]

Avantol13:


suites/apis/googleServiceAccountTest.js
```diff
@@ -347,55 +402,56 @@ Scenario('Register Google Service Account for commons project without privilege
...
+  // Find a user's google email meeting following conditions
+  //   -an email that is NOT the owner of the google project
+  //   -an email that is NOT the email of the userWithoutPrivilege
+  //   -an email that is NOT the email of the userWithPrivilege
+  // It is assumed that all users' usernames are google account emails
+  const differentGoogleEmail = Object.values(users).find(user =>
+    user.username !== googleProject.owner &&
+    user.username !== userWithoutPrivilege.username &&
+    user.username !== userWithPrivilege.username,
+  ).username;
```
macintoshpie: I wrote this hacky code - I don't really like how I did this previously but currently we don't have something implemented to find this easily. I would just add this extra bit to make it crash and burn if we don't find an email to use.
```suggestion
  ).username;
  if (differentGoogleEmail === undefined) {
    throw "Unable to find a user to add to Google project";
  }
```

PR #82, opened 2019-01-30
fix(testfilter): grep/invert confusion

frickjack:
* more error checking
* increase timeout
* delete dead code

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -89,8 +89,8 @@ lockUser=""
 
 testArgs="--debug --verbose --reporter mocha-junit-reporter"
 #
-# DCF Google tests are not yet stable enough to run in fence PR's -
-# so just enable in gen3-qa PR's for now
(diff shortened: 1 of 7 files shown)


PlanXCyborg:
Jenkins Build 1 : time taken 30 min

run-tests.sh
```diff
@@ -89,13 +89,16 @@ lockUser=""
 
 testArgs="--debug --verbose --reporter mocha-junit-reporter"
 #
-# DCF tests are not yet stable enough to run in fence PR's -
+# DCF Google tests are not yet stable enough to run in fence PR's -
```
thanh-nguyen-dang: you said it is not stable to run in fence? But I see it will run when `$service == "fence"` line 99 and 103

PlanXCyborg:
Jenkins Build 2 : time taken 4 min 37 sec

PR #83, opened 2019-01-30
Fix/servicecheck

frickjack:
add service checks to run-tests.sh


PlanXCyborg:
Jenkins Build 1 : time taken 22 min

PR #84, opened 2019-01-31
fix(s3cleanup): leave breadcrumbs in homedir

frickjack:


PlanXCyborg:
Jenkins Build 1 : time taken 29 min

PR #91, opened 2019-02-07
Fix/login

paulineribeyre:
- windmill tests log in via the login button instead of by manually setting a cookie
- use default fake creds for DCF tests instead of throwing an error if they are not set in env vars

### DIFF
--- a/utils/user.js
+++ b/utils/user.js
@@ -53,7 +53,7 @@ class User {
    * @returns {{email: string, password: string}}
    */
   get googleCreds() {
-    if (!process.env[this.envGoogleEmail] || process.env[this.envGoogleEmail] === '' || !process.env[this.envGooglePassword] || process.env[this.envGooglePassword] === '') {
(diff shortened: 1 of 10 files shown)


utils/user.js
```diff
@@ -53,8 +53,9 @@ class User {
    * @returns {{email: string, password: string}}
    */
   get googleCreds() {
-    if (process.env[this.envGoogleEmail] === '' || process.env[this.envGooglePassword] === '') {
-      throw Error(GOOGLE_CREDS_MISSING_ERROR + this.username);
+    if (!process.env[this.envGoogleEmail] || process.env[this.envGoogleEmail] === '' || !process.env[this.envGooglePassword] || process.env[this.envGooglePassword] === '') {
```
Avantol13: why are the additional `!process.env[this.envGoogleEmail]` necessary? doesn't `''` evaluate to false? or am I thinking too much like Python..

paulineribeyre:
jenkins didn't pass because this PR requires changes from https://github.com/uc-cdis/fence/pull/555 -> merging both at the same time

PR #93, opened 2019-02-12
fix(login-logout): remove default login; add logout logic

paulineribeyre:
Fixes a bug where a test logs in, doesn't log out, and the following tests fail because they don't find the login button
- do not set an `access_token` to log in at the beginning of the tests
- log in and log out using the portal buttons before and after each windmill test

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ You can kill the server with `npm run selenium-kill`
 ## Running Tests
 Once you have your environment variables configured and the Selenium server is running, you should be able to successfully run the tests.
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -106,10 +106,12 @@ You can kill the server with `npm run selenium-kill`
 ## Running Tests
 Once you have your environment variables configured and the Selenium server is running, you should be able to successfully run the tests.
 
-If you want to run `gen3-qa` against a dev environment, you just need to set the environment variable `NAMESPACE={dev env namespace}` and the `GOOGLE_APP_CREDS_JSON` to the credentials getting from Google, then run `./run-tests-localsh`
+If you want to run `gen3-qa` against a dev environment, you just need to set the environment variable `NAMESPACE={dev env namespace}` and the `GOOGLE_APP_CREDS_JSON` to the credentials getting from Google, then run `./run-tests-localsh`.
```
qingyashu: typo: `./run-tests-local.sh`  :p

PR #97, opened 2019-02-26
DCF expiration and removal tests

paulineribeyre:
Add expiration and removal DCF integration tests:
- Locking setup, documentation and test
- SA registration expiration test
- SA temporary key expiration test
- SA validation (`google-manage-user-registrations` job) use cases

Use cases I did not implement:
- Link google account expiration: there is no way to check whether the expiration worked
- SA validation - check SA policy is accessible: same use case as monitor does not have access
- SA validation - check monitor account exists: when re-creating the account, we would need to set up fence with the new SA key
- SA validation - check the user is in the Google project; check the other users of the project are valid; check the other SAs of the project are valid: these are not relevant to the validation job
- SA validation - check valid parent organization: there is no way right now to programmatically update the whitelist
- SA validation - check projects exists: the google cloud API does not let us create/delete projects
- SA validation - check SA is owned by the Google project; check valid type of SA: not relevant to the validation job because we would not be able to register the SA in the first place

Other changes:
- Add debug logs and retry when timeout to `registerGoogleServiceAccount()`
- Add retry to flaky Google tests and ETL tests
- Refactor utils (deleted unused `CommonsUtil` and renamed `dataUploadUtil`)
- The tests now use `gen3qa-service@gen3qa-validationjobtest.iam.gserviceaccount.com` instead of `fence-service@dcf-integration.iam.gserviceaccount.com`, because `fence-service` was both the SA used by the tests and the monitor SA, and we need to separate these for testing purposes (updated creds in the Jenkins envs)


services/apis/fence/fenceQuestions.js
```diff
@@ -128,7 +131,12 @@ module.exports = {
    * @param {Gen3Response} expectedRes
    */
   responsesEqual(actualRes, expectedRes) {
-    expect(actualRes).to.be.a.gen3Res(expectedRes);
+    err = '';
+    try {
```
Avantol13: why was this necessary?

paulineribeyre: it was helpful to know what went wrong when a service account registration failed, but later i added a log directly in `registerGoogleServiceAccount`
so i can remove this

utils/google.js
```diff
@@ -254,4 +309,165 @@ module.exports = {
...
+    isLocked = listRes.some(sa => sa.email.startsWith(lockServiceAccountName));
+    if (!isLocked) return true;
+
+    // if it's locked by another testing session: cannot unlock
+    isLockedByMe = listRes.some(sa => sa.email === serviceAccountEmail);
+    if (!isLockedByMe) return false;
+
+    // if it's lock by this testing session: try to unlock
+    console.log(`Unlocking Google project "${googleProject.id}"`);
+    const deleteRes = await this.deleteServiceAccount(serviceAccountEmail, googleProject.id);
```
Avantol13: does this work pretty much every time or should you retry this if it fails?

services/apis/fence/fenceSequences.js
```diff
@@ -112,5 +113,60 @@ module.exports = {
...
+
+
+  /**
+   * Cleans up fence's DBs for links and service accounts
+   * Takes the google and users utils as params
+   * @returns {Promise<void>}
+   */
+  async suiteCleanup(google, users) {
+    // unlock the lockable google project
+    await google.unlockGoogleProject(fenceProps.googleProjectDynamic);
```
Avantol13: shouldn't you get the response from `unlockGoogleProject` and make sure it's true?

suites/apis/googleServiceAccountRemovalTest.js
```diff
@@ -0,0 +1,471 @@
...
+});
+
+
+After(async (google, fence, users) => {
+  await fence.complete.suiteCleanup(google, users);
+});
+
+
+AfterSuite(async (google, fence) => {
+  await google.unlockGoogleProject(fence.props.googleProjectDynamic);
```
Avantol13: shouldn't you get the response from unlockGoogleProject and make sure it's true?

services/apis/fence/fenceSequences.js
```diff
@@ -112,5 +113,60 @@ module.exports = {
...
+
+    // google projects to 'clean up'
+    const googleProjects = [
+      fenceProps.googleProjectA,
```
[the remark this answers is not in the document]
paulineribeyre: `suiteCleanup` already includes ProjectA and ProjectWithComputeServiceAcct right now, I just added ProjectDynamic - from what I understand it cleans up the DB, in case a test crashed and didn't clean up

suites/apis/googleServiceAccountRemovalTest.js
```diff
@@ -0,0 +1,471 @@
...
+
+  // Access data
+  user0AccessQAResAfter = await google.getFileFromBucket(
+    fence.props.googleBucketInfo.QA.googleProjectId,
+    pathToKeyFile,
+    fence.props.googleBucketInfo.QA.bucketId,
+    fence.props.googleBucketInfo.QA.fileName
+  );
+
+  // Clean up
```
Avantol13: this won't happen if this fails before this on the `fence.ask` or `expect` for locking. will that cause an issue for a later test env? the SA's are per env now, so that shouldn't matter. but I'm worried about the linked google accounts which are shared. are you sure that every test removes all linked accounts before trying anything?  Or are you sure that already having an account linked won't break other tests?

suites/apis/googleServiceAccountRemovalTest.js
```diff
@@ -0,0 +1,471 @@
...
+
+  chai.expect(user0AccessQAResAfter,
+    'User should have bucket access after clean up job'
+  ).to.have.property('id');
+
+  chai.expect(unlockRes, 'Could not unlock project').to.be.true;
+}).retry(2);
+
+
+// TODO: enable. this test fails until the jenkins envs are setup with the new monitoring SA
```
Avantol13: remove this line?
```suggestion
```

suites/apis/googleServiceAccountRemovalTest.js
```diff
@@ -0,0 +1,471 @@
...
+xScenario('SA removal job test: monitor SA does not have access @reqGoogle', async (fence, users, google, files) => {
```
Avantol13:
```suggestion
Scenario('SA removal job test: monitor SA does not have access @reqGoogle', async (fence, users, google, files) => {
```

suites/apis/googleServiceAccountRemovalTest.js
Avantol13: Oh, wait, do the jenkins envs have the new SA? We can do that first and make sure it doesn't break existing tests

suites/apis/googleServiceAccountRemovalTest.js
```diff
@@ -0,0 +1,471 @@
...
+  // test invalid SA because the monitor does not have access
+
+  // Setup
+  const googleProject = fence.props.googleProjectDynamic;
+  let getMonitorRes = await fence.do.getGoogleServiceAccountMonitor(users.user0);
+  fence.ask.assertStatusCode(getMonitorRes, 200, 'Could not get SA monitor info');
+  let monitorEmail = getMonitorRes.body.service_account_email;
```
Avantol13: this could fail if the monitoring endpoint changes what it returns (i.e. no `service_account_email` field). should make sure the tests respond with a sensical error in that case

suites/apis/googleServiceAccountRemovalTest.js
```diff
@@ -0,0 +1,471 @@
...
+  await fence.complete.suiteCleanup(google, users);
+});
+
+
+AfterSuite(async (google, fence) => {
+  await google.unlockGoogleProject(fence.props.googleProjectDynamic);
+});
+
+
+function checkAndCleanSA() {
```
Avantol13: can we rename this to something like `runVerifyUserSAsJob` or something similar?

suites/apis/googleServiceAccountRemovalTest.js
```diff
@@ -0,0 +1,471 @@
...
+    fence.props.googleBucketInfo.QA.bucketId,
+    fence.props.googleBucketInfo.QA.fileName
+  );
+
+  // Clean up
+  console.log('cleaning up');
+
+  // Add monitor's access back
+  for (var role of monitorRoles) {
+    await google.updateUserRole(
```
Avantol13: what if this fails? we should probably check the response and provide help if it fails.. b/c someone will have to manually re-add it, right? probably safe to try and add it add the beginning and catch any conflict errors and ignore/continue. that way hopefully no one has to manually touch these projects even if this fails

suites/apis/googleServiceAccountRemovalTest.js
Avantol13: so the db won't have the google account linking information in another env b/c we wipe the db. but the actual google account may still be in the proxy group.. I'm trying to think through if this is a problem

Avantol13: actually, I think this is fine... b/c when we do SA validation, to check linkage we check the fence db *not* Google. and we don't have any tests for accessing data with the linked account

suites/apis/googleServiceAccountRemovalTest.js
```diff
@@ -0,0 +1,471 @@
...
+    users.user0,
+    googleProject.serviceAccountEmail,
+  );
+
+  await fence.do.unlinkGoogleAcct(
+    users.user0,
+  );
+
+  // Unlock the project
+  let unlockRes = await google.unlockGoogleProject(googleProject);
```
Avantol13: what if this fails :grimacing: ? we probably need a big error to the user about how to manually remove the lock

suites/apis/googleServiceAccountRemovalTest.js
```diff
@@ -0,0 +1,471 @@
...
+  registerRes = await fence.do.registerGoogleServiceAccount(
+    users.user0,
+    googleProject,
+    ['QA']
+  );
+  fence.ask.responsesEqual(registerRes, fence.props.resRegisterServiceAccountSuccess);
+
+  // Lock the project
+  lockRes = await google.lockGoogleProject(googleProject);
+  chai.expect(lockRes, 'Could not lock project').to.be.true;
```
Avantol13: can this be a bit more specific of an error message in case someone isn't familiar with what it tried to do and how it could fail? I'd rather have a paragraph here explaining then have someone bug you later about what this means. so like: `Could not lock Google project to run tests where we modifying the project iteself. Locks are necessary to avoid interfering with other testing envs since they all use this Google Project: {{name}}. We attempted to "lock" the env by creating a service account unique to this test run. We either could not create the SA or {{insert other potential failures here}}, the tests tried multiple times. Perhaps check the logs for Google API responses.`

Avantol13: I know it seems excessive but it'll really help someone in the future if this randomly doesn't work

Avantol13: something similar for `unlock` failures too.. possibly with more info about how to remove the stray locking SA if it fails

suites/apis/googleServiceAccountRemovalTest.js
```diff
@@ -0,0 +1,471 @@
...
+  let registerRes = await fence.do.registerGoogleServiceAccount(
+    users.user0,
+    googleProject,
+    ['QA']
+  );
+  fence.ask.responsesEqual(registerRes, fence.props.resRegisterServiceAccountSuccess);
+
+  // Lock the project
+  lockRes = await google.lockGoogleProject(googleProject);
+  chai.expect(lockRes, 'Could not lock project').to.be.true;
```
Avantol13: more details, see prev comment

suites/apis/googleServiceAccountRemovalTest.js
```diff
@@ -0,0 +1,471 @@
...
+  let unlockRes = await google.unlockGoogleProject(googleProject);
+
+  // Asserts
+  fence.ask.detected_invalid_google_project(jobRes, fence.props.monitorSAJobLog.noFenceUser);
+
+  chai.expect(user0AccessQARes,
+    'User should NOT have bucket access after clean up job'
+  ).to.have.property('statusCode', 403);
+
+  chai.expect(unlockRes, 'Could not unlock project').to.be.true;
```
Avantol13: more details, see prev comment

suites/apis/googleServiceAccountRemovalTest.js
```diff
@@ -0,0 +1,471 @@
...
+
+  // Register account
+  let registerRes = await fence.do.registerGoogleServiceAccount(
+    users.user0,
+    googleProject,
+    ['QA']
+  );
+  fence.ask.responsesEqual(registerRes, fence.props.resRegisterServiceAccountSuccess);
+
+  // Lock the project
```
Avantol13: we may want to lock before registering the SA. Fence does a ton of checks against the project to make sure the SA can be registered, if it's already locked and invalid, registration may fail

suites/apis/googleServiceAccountRemovalTest.js
```diff
@@ -0,0 +1,471 @@
...
+  );
+
+  // Run clean up job
+  let jobRes = checkAndCleanSA();
+
+  // Make sure we cannot access data after the clean up
+  /**
+   * return true if the user can access data, false otherwise
+   */
+  const isDataUnaccessible = async function() {
```
Avantol13:
```suggestion
  const isDataInaccessible = async function() {
```

suites/apis/googleServiceAccountRemovalTest.js
```diff
@@ -0,0 +1,471 @@
...
+    }
+    catch {
+      console.log('Data is still accessible: rerunning');
+      return false;
+    }
+  };
+  const timeout = 60; // max number of seconds to wait
+  let user0CanAccessQAResAfter = false;
+  try {
+    await apiUtil.smartWait(isDataUnaccessible, [], timeout, '');
```
Avantol13:
```suggestion
    await apiUtil.smartWait(isDataInaccessible, [], timeout, '');
```

suites/apis/googleServiceAccountTest.js
```diff
@@ -100,6 +59,57 @@ Scenario('Register Google Service Account Success @reqGoogle @first', async (fen
   fence.ask.responsesEqual(deleteRes, fence.props.resDeleteServiceAccountSuccess);
 });
 
+Scenario('Google project locking test @reqGoogle', async (fence, google) => {
```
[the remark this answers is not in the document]
Avantol13: does this run before everything else?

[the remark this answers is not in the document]
Avantol13: does this test take a long time to run? with retries and whatnot?

utils/commonsUtil.js
```diff
@@ -1,138 +0,0 @@
-/**
```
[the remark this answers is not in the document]
paulineribeyre: it got renamed to `commons.js` during a refactor but the old file never got deleted

suites/apis/googleServiceAccountTest.js
```diff
@@ -547,5 +564,171 @@ Scenario('Delete a SA that was successfully registered before but was deleted fr
...
+  );
+
+  // Wait for the link to expire
+  console.log('waiting for the link to expire');
+  await apiUtil.sleepMS(EXPIRES_IN * 1000);
+
+  // Run the expired SA clean up job
+  console.log('Clean up expired Service Accounts');
+  bash.runJob('google-delete-expired-service-account-job');
+  await apiUtil.sleepMS(5 * 1000); // wait for propagation to google
```
Avantol13: can you do a smart wait here instead of just strict wait?

Avantol13: might save us a few seconds

utils/google.js
```diff
@@ -206,13 +218,54 @@ module.exports = {
...
+    saList = [];
+    nextPageToken = '';
+    while (typeof nextPageToken !== 'undefined') {
+      listRes = await new Promise((resolve) => {
+        return googleApp.authorize(googleApp.cloudManagerConfig, (authClient) => {
+          const cloudResourceManager = google.iam('v1');
+          const request = {
+            name: `projects/${projectID}`,
+            auth: authClient,
+            pageSize: new Number(5), // get the first 5 SAs (this does not work well)
```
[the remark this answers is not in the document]
paulineribeyre: sometimes it will just return 1 or 2 items even though there are more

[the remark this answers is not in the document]
paulineribeyre: that's why i had to use the nextPageToken

utils/google.js
paulineribeyre: i have never seen it fail

suites/apis/googleServiceAccountRemovalTest.js
paulineribeyre: i will change this to lock before registering the SA like you suggested, so if `expect(lock)` fails there will be no SA creation so nothing to delete. for the linked google account, it should be unlinked by the suite clean up

suites/apis/googleServiceAccountRemovalTest.js
paulineribeyre: yeah it's not setup yet

suites/apis/googleServiceAccountRemovalTest.js
paulineribeyre: we can stop skipping this test once jenkins envs are set up with the new SA

suites/apis/googleServiceAccountTest.js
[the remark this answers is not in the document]
paulineribeyre: it does not run first :(
it runs first in the `googleServiceAccountTest` suite, but from the logs it looks like `googleServiceAccountRemovalTest` runs before `googleServiceAccountTest`. but it will still help to understand what's wrong if we see it fail
it took 80 sec just now on my dev env

suites/apis/googleServiceAccountRemovalTest.js
paulineribeyre: i added a second role update in the `after suite` to make sure it gets done - we can't just add it everywhere because we need to lock the project to updates roles, and that's messy
i will also move this test to run last in the suite, so that a failure to update the roles doesn't make the following tests fail

services/apis/fence/fenceSequences.js
paulineribeyre: actually - we can't assert here, since we don't want it to throw if the lock belongs to another session

suites/apis/googleServiceAccountRemovalTest.js
paulineribeyre: done in 61b67026498d7a8503fc5d0e85c47c242ed5ac10

PR #104, opened 2019-03-12
DCF google linking tests

paulineribeyre:
Add Google linking endpoint tests:
- linking an account that is already linked to this user should fail
- linking an account that is already linked to different user should fail
- extending a link's expiration after it expired should succeed

Not implemented because Google limits logging in with captchas:
- link google account A and verify you can access data with the linked google account
- unlink google account A and link google account B - verify you can’t access data using googleA, and you can access data using googleB


suites/apis/linkGoogleAccountTest.js
```diff
@@ -39,3 +67,23 @@ Scenario('try to extend link when acct is not linked @reqGoogle', async (fence,
   const extendRes = await fence.do.extendGoogleLink(users.mainAcct);
   fence.ask.responsesEqual(extendRes, fence.props.resExtendNoGoogleAcctLinked);
 });
+
+Scenario('link an acct that is already linked to this user @reqGoogle', async (fence, users) => {
+  await fence.complete.linkGoogleAcctMocked(users.mainAcct);
+
+  // try to link mainAcct to the same account again
+  let linkRes = await fence.do.linkGoogleAcctMocked(users.mainAcct, users.mainAcct.googleCreds.email);
+  expect(linkRes.finalURL, 'Linking a google account twice should fail').to.not.exist;
```
Avantol13: this should actually redirect with an `error` and `error_description`.. we should be able to check the `error` here... what is the response? Why is finalURL not getting filled out?

Avantol13: should look like: `https://{{domain}}/?error_description=User+already+has+a+linked+Google+account.&error=g_acnt_link_error`

paulineribeyre: this error is in the url in `linkGoogleAcctMocked`, but when we GET it, there is no `location` in the response headers.
actually i don't understand why we GET so many times (https://github.com/uc-cdis/gen3-qa/blob/master/services/apis/fence/fenceTasks.js#L219-L220), do you know why?

Avantol13: wait, so you do see the error in the url? can we check for that here?

paulineribeyre: fixed in 73ede732b60574df689d06e8a89137314038229e

services/apis/fence/fenceTasks.js
```diff
@@ -218,7 +218,7 @@ module.exports = {
     return I.sendGetRequest(url, headers).then((res) => {
       // follow redirect back to fence
       let sessionCookie = getCookie('fence', res.headers['set-cookie']);
-      headers = { Cookie: 'dev_login=' + mockLoginUser + ';' + 'fence=' + sessionCookie };
+      headers.Cookie += `; fence=${sessionCookie}`;
```
Avantol13: wait, don't we need the dev_login cookie for other tests?

paulineribeyre: the headers already contain access_token and dev_login, the fix is appending the session cookie (before the fix, this line was removing the access_token from the headers)

PR #106, opened 2019-03-15
Feat/dcf other endpoints

paulineribeyre:
Created a new test suite "googleServiceAccountKey" added DCF tests:
- Get current service accounts creds and check their expiration time
- No access to data after the SA key is deleted
- Deleting SA creds that do not exist returns 404
- Expired SA creds that exist in fence but do not exist in Google should get deleted by the `google-manage-keys` job
- The project access in the ID token is the same as the project access in the fence user endpoint


suites/apis/googleServiceAccountKeyTest.js
```diff
@@ -0,0 +1,349 @@
...
+
+  key2 = credsList3.filter(key => key.name.includes(keyId2));
+  expect(key2.length,
+    'Only the SA key that was not deleted should be listed'
+  ).to.equal(1);
+});
+
+
+Scenario('Get temporary Service Account creds, get object in bucket, delete creds @reqGoogle', async (fence, users, google, files) => {
+  // Test that we do not have access to data anymore after the SA key is deleted
```
Avantol13: can this be the scenario text? and the scenario text with details of what happens can be a comment?

suites/apis/googleServiceAccountKeyTest.js
```diff
@@ -0,0 +1,349 @@
...
+
+  const googleProject = fence.props.googleProjectA;
+
+  // Setup
+  await fence.complete.forceLinkGoogleAcct(users.user0, googleProject.owner);
+
+  // Register account
+  const registerRes = await fence.do.registerGoogleServiceAccount(
```
Avantol13: I don't think we need this. registering is only necessary for User SA Data Access, but is not needed for Temporary Creds

suites/apis/googleServiceAccountKeyTest.js
```diff
@@ -0,0 +1,349 @@
...
+});
+
+
+Scenario('Get temporary Service Account creds, get object in bucket, delete creds @reqGoogle', async (fence, users, google, files) => {
+  // Test that we do not have access to data anymore after the SA key is deleted
+
+  const googleProject = fence.props.googleProjectA;
+
+  // Setup
+  await fence.complete.forceLinkGoogleAcct(users.user0, googleProject.owner);
```
Avantol13: we don't need to link for temp creds access, this is only for User SA data access

suites/apis/googleServiceAccountKeyTest.js
```diff
@@ -0,0 +1,349 @@
...
+
+
+Scenario('SA key removal job test: remove expired creds @reqGoogle', async (fence, users, google, files) => {
+  // Test that we do not have access to data anymore after the SA key is expired
+
+  const EXPIRES_IN = 1;
+  const googleProject = fence.props.googleProjectA;
+
+  // Setup
+  await fence.complete.forceLinkGoogleAcct(users.user0, googleProject.owner);
```
Avantol13: same as above, don't think we need this linking and sa registration

suites/apis/googleServiceAccountKeyTest.js
```diff
@@ -0,0 +1,349 @@
...
+
+
+Scenario('SA key removal job test: remove expired creds that do not exist in google @reqGoogle', async (fence, users, google) => {
+  // Test that the job removes keys from the fence DB even if some of them do not exist in google
+
+  const EXPIRES_IN = 1;
+  const googleProject = fence.props.googleProjectA;
+
+  // Setup
+  await fence.complete.forceLinkGoogleAcct(users.user0, googleProject.owner);
```
Avantol13: same as above, don't think this is needed

test_setup.js
```diff
@@ -183,9 +183,10 @@ module.exports = async function (done) {
       process.env[user.envTokenName] = at;
     }
 
+    const mainAcct = users.mainAcct;
     console.log('Delete then create basic client...\n');
     deleteClient('basic-test-client');
-    const basicClient = createClient('basic-test-client', 'test-client@example.com', 'basic');
+    const basicClient = createClient('basic-test-client', mainAcct.username, 'basic');
```
Avantol13: was this breaking something?

paulineribeyre: i thought i needed that but i actually don't! haha i'll remove it

suites/apis/googleServiceAccountKeyTest.js
paulineribeyre: 😮

paulineribeyre:
bug fixes in 2e3d03c5c1ad7b485d278c35aee6638ecbe9f9d1:
- test failure when monitor access is revoked in the middle of the job running
- afterAll failure when another jenkins session locks the project and we try to unlock it and check for success

PR #114, opened 2019-04-02
feat(rbac): centralized auth integration tests

Avantol13:
Integration testing for future Arborist changes

> NOTE: This relies on https://github.com/uc-cdis/commons-users/pull/399/

## Test Plan

1) - Test that user without policies can't CRUD records in /gen3 or /abc

   - Test user with `abc-admin` policy, test creating record, reading record, updating
     record, and deleting in indexD with rbac /abc (should work),
     make sure they still can't create, read, update OR delete in /gen3

   - Test the same thing as above, but use a client's access_token for that user

2) - Test CLIENTB creating signed urls for data under /abc, ensure they cannot create
     signed urls for data under /gen3. Ensure that successful signed urls actually
     allow reading data.

   - Test client WITH access creating signed urls for data for user WITHOUT access in namespace.

   - Test client WITHOUT access creating signed urls for data for user WITH access in namespace.

   - Test that a user with `abc-admin` can create signed urls with their own access_token

3) - Test client flow to get id_token, compare to what's in userinfo endpoint, ensure
     RBAC policy information is there
        > NOTE: there is an existing test that might satisfy this with a few updates

4) - Create some records in indexd with AND logic, have CLIENTB try to create signed
     url for file for user that does NOT have necessary permissions, ensure failure.

   - Same as above but user DOES have permissions. Ensure successful signed URL that
     we can access data with


services/apis/indexd/indexdTasks.js
```diff
@@ -82,31 +86,77 @@ module.exports = {
...
+   */
+  async updateFile(file, data, headers = user.mainAcct.accessTokenHeader) {
+    return I.sendGetRequest(
+      `${indexdProps.endpoints.get}/${file.did}`,
+      headers,
+    ).then((res) => {
+      // get last revision
+      file.rev = getRevFromResponse(res);
+
+      const headers = user.mainAcct.indexdAuthHeader;
```
paulineribeyre: why do you use `mainAcct`? does `indexdAuthHeader` require the user to be an admin? if so, could you add a comment to explain this (either at `indexdAuthHeader` or here or both)

services/apis/indexd/indexdTasks.js
```diff
@@ -82,31 +86,77 @@ module.exports = {
...
    * @param {Object} file
    * @returns {Promise<Promise|*|PromiseLike<T>|Promise<T>>}
    */
-  async deleteFile(file) {
-    return I.sendDeleteRequest(
-      `${indexdProps.endpoints.delete}/${file.did}?rev=${file.rev}`,
-      user.mainAcct.indexdAuthHeader,
+  async deleteFile(file, headers = user.mainAcct.indexdAuthHeader) {
+    // if we already have the current revision we can use it, otherwise we need to get it
+    if (file.rev !== null && file.rev !== undefined) {
```
paulineribeyre: why not something like this?
```
if (!file.rev) {
    file.rev = await I.sendGetRequest(...).then(res => getRevFromResponse(res));
}
return I.sendDeleteRequest(...);
```

services/apis/fence/fenceProps.js
```diff
@@ -422,6 +431,9 @@ module.exports = {
     client: new Client({
       envVarsName: 'CLIENT',
     }),
+    clientb: new Client({
```
paulineribeyre: why do you need a new basic client?
and i realize this is nitpicky but could you rename `clientImplicit.envVarsName` to not be `CLIENTB` for clarity?

utils/apiUtil.js
```diff
@@ -113,6 +113,18 @@ const gen3Res = function (_chai, _) {
 };
 
 module.exports = {
+  /**
+   * Returns the JSON for an access token header given the token itself
+   * @param {string} accessToken - access token string
+   * @returns {JSON}
+   */
+  getAccessTokenHeader(accessToken) {
```
paulineribeyre: awesome!! by searching "`bearer" in the repo i find a bunch of places we could use this :p if you feel like refactoring stuff :p

suites/apis/centralizedAuth.js
```diff
@@ -0,0 +1,655 @@
...
+
+/*
+   - Test the same thing as above, but use a client's access_token for that user
+*/
+Scenario('Client (with access) with user token (with access) can CRUD indexd records in namespace, not outside namespace @rbac',
+  async (fence, indexd, users, files) => {
+    return; // FIXME: skip test for now
+    // NOTE: default CLIENT is abc-admin and gen3-admin
+    const tokenRes = await fence.complete.getUserTokensWithClient(users.mainAcct);
+    const accessToken = tokenRes.body.access_token;
```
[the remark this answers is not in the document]
Avantol13: :+1: will use `asssertTokensSuccess`

services/apis/indexd/indexdTasks.js
Avantol13: ah, this is a mistake, should use indexdAuthHeader by default and allow overriding to user token. will fix. auth header is just basic auth with username/password and doesn't actually rely on a specific user (it just pulls from env vars).

services/apis/indexd/indexdTasks.js
Avantol13: :+1:

utils/apiUtil.js
Avantol13: :+1:

services/apis/fence/fenceProps.js
Avantol13: For the tests, I want to configure the clients to have different access and be able to test access/denied. You can checkout the comments in centralizedAuth.js and also the common-users PR with the arborist block for more info on the diff permissions. I could rename to `anotherBasicClient` or something, or `basicClient2` if you think that'd improve clarity

suites/apis/centralizedAuth.js
```diff
@@ -0,0 +1,655 @@
...
+const bash = new Bash();
+
+const fs = require('fs');
+
+const I = actor();
+
+/*
+for twoProjectsFile, use cdis.autotest@gmail.com
+
+Each has ONE of these policies:
```
paulineribeyre: i find this comment a bit confusing, does it mean "for twoProjectsFile: cdis.autotest@gmail.com has access to both projects, dummy-one and smarty-two have access to one each"?

suites/apis/centralizedAuth.js
```diff
@@ -0,0 +1,655 @@
...
+let new_gen3_records = {
+  fooBarFile: {
+    filename: 'testdata',
+    link: 's3://cdis-presigned-url-test/testdata',
+    md5: '73d643ec3f4beb9020eef0beed440ad4',
+    acl: [],
+    rbac: ['/gen3/programs/test/projects/test'],
+    size: 9,
+  },
+  deleteMe: {
```
paulineribeyre: for indexd DELETE, you only need `did` and `rev`:
- i think you don't need to keep so much information in your `deleteMe` files
- maybe i missed it but i don't see where you create the `deleteMe` files in indexd, and i don't think deletion tests will work if you don't create the records first... you should use `addFileIndices` on these files in the `Before` block, it will populate `file.rev`

suites/apis/centralizedAuth.js
```diff
@@ -0,0 +1,655 @@
...
+    const abc_update_response = indexd.do.updateFile(
+        new_abc_records.fooBarFile, filename_change, users.user2.accessTokenHeader)
+
+    // delete
+    const gen3_delete_response = indexd.do.deleteFile(
+        new_gen3_records.deleteMe, users.user2.accessTokenHeader)
+    const abc_delete_response = indexd.do.deleteFile(
+        new_abc_records.deleteMe, users.user2.accessTokenHeader)
+
+    // asserts for create
```
paulineribeyre: for these tests i think it would be safe to do the asserts right after the tests

suites/apis/centralizedAuth.js
```diff
@@ -0,0 +1,655 @@
...
+Scenario('Client (with access) with user token (with access) can CRUD indexd records in namespace, not outside namespace @centralizedAuth',
+  async (fence, indexd, users, files) => {
+    return; // FIXME: skip test for now
+    // NOTE: default CLIENT is abc-admin and gen3-admin
+    const tokenRes = await fence.complete.getUserTokensWithClient(users.mainAcct);
+    const accessToken = tokenRes.body.access_token;
+
+    // create
+    const gen3_create_success = await indexd.do.addFileIndices(
+        Object.values(new_gen3_records), apiUtil.getAccessTokenHeader(users.mainAcct.accessToken));
```
paulineribeyre:
```suggestion
        Object.values(new_gen3_records), apiUtil.getAccessTokenHeader(accessToken));
```
this typo is in several tests

services/apis/fence/fenceProps.js
paulineribeyre: oh sorry, i didn't notice the clients were defined in the user.yaml so i thought they had the same access
for the renaming, i meant this line https://github.com/uc-cdis/gen3-qa/blob/48b92f4ca0b2806332a099f2ac395fe2fd017f98/services/apis/fence/fenceProps.js#L442

paulineribeyre: you could also add comments in here about which client has which policies in the user.yaml, similar to what we did for users https://github.com/uc-cdis/gen3-qa/blob/master/utils/user.js#L105

Avantol13: oh, yes, good idea! and I just relooked at this, I did mess up and put clientb for ABC_CLIENT. I will fix

services/apis/indexd/indexdTasks.js
```diff
@@ -82,27 +86,71 @@ module.exports = {
...
       return res.body;
     });
   },
 
+  /**
+   * Updates indexd data for file
+   * @param {Object} fileNode - Assumed to have a did property
+   * @returns {Promise<Gen3Response>}
+   */
+  async updateFile(file, data, headers = user.mainAcct.indexdAuthHeader) {
```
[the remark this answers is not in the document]
Avantol13: this is backwards compatible in that it will default to sending auth header both times (which should be fine for GET in old indexd). then in new indexd, it needs access token for both, so we just pass in the headers arg in those tests

suites/apis/centralizedAuth.js
Avantol13: yeah, this is sorta outdated. I will remove since I added details about access in the comments above

suites/apis/centralizedAuth.js
Avantol13: these get created in the test itself: https://github.com/uc-cdis/gen3-qa/blob/5ea07c7d23e5b4143091d481a182ee4c18f3cf0b/suites/apis/centralizedAuth.js#L213 when it adds the other files too

services/apis/indexd/indexdTasks.js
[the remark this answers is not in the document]
paulineribeyre: oh i see! let's rename the param `authHeaders` to make it clear then? and change which headers you pass to this function in the tests

suites/apis/centralizedAuth.js
Avantol13: oh! shoot, good catch

suites/apis/centralizedAuth.js
Avantol13: :+1:

services/apis/indexd/indexdTasks.js
paulineribeyre: you shouldn't need to duplicate the `sendDeleteRequest` block

Avantol13: ah, whoops, will fix

suites/apis/centralizedAuth.js
```diff
@@ -0,0 +1,639 @@
...
+        new_gen3_records.deleteMe, users.mainAcct.accessTokenHeader)
+    const abc_delete_response = indexd.do.deleteFile(
+        new_abc_records.deleteMe, users.mainAcct.accessTokenHeader)
+
+    // asserts for delete
+    fenceQuestions.assertStatusCode(
+        gen3_delete_response, 403,
+        msg='should have gotten unauthorized for deleting record under `/gen3`'
+    );
+    fenceQuestions.assertStatusCode(
```
paulineribeyre: there is a `deleteFileSuccess` function in `indexdQuestions` that you could use instead of checking code=200. I think adding the code=200 check to that function would be nice, and if you want to be thorough you can also call `getFile` and check that no file is found (also in the `deleteFileSuccess` function)

suites/apis/centralizedAuth.js
```diff
@@ -0,0 +1,639 @@
...
+        abc_read_response, 403,
+        msg='should have gotten unauthorized for reading record under `/abc`'
+    );
+
+    // update
+    const filename_change = {
+      'file_name': 'test_filename'
+    }
+    const gen3_update_response = indexd.do.updateFile(
+        new_gen3_records.fooBarFile, filename_change, users.user2.accessTokenHeader)
```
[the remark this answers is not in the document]
paulineribeyre: btw i am curious/confused about why these tests passed when you ran them with `accessTokenHeader`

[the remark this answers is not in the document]
Avantol13: :thinking: hmm... I will need to investigate this one. Now that you point it out, I am also confused

services/apis/fence/fenceTasks.js
```diff
@@ -111,7 +111,7 @@ module.exports = {
     return I.sendGetRequest(url).then(res => res.body);
   },
 
-  getGoogleFileFromSignedUrlRes(signedUrlRes) {
```
paulineribeyre: i think this function would be better as a 2-step fence **sequence** (similar to `checkFileEquals()`)
- `fenceQuestions.hasUrl(signedUrlRes, FILE_FROM_URL_ERROR)` (you can add optional error message param to `hasUrl`)
- `return await fenceTasks.getFile(signedUrlRes.body.url)`

Avantol13: This isn't getting a signed url from fence though, it's actually taking a signed URL response as input, and following the URL to get the actual file contents from cloud storage

paulineribeyre: but what's the difference? the `if` block checks the same thing `hasUrl` does (i guess you can change the error message to make it more general if the url is not from fence) and the `get` does the same thing as `getFile(signedUrlRes.body.url)`

services/apis/indexd/indexdTasks.js
```diff
@@ -119,6 +119,7 @@ module.exports = {
    * @returns {Promise<Gen3Response>}
    */
   async updateFile(file, data, authHeaders = user.mainAcct.indexdAuthHeader) {
+    authHeaders['Content-Type'] = 'application/json; charset=UTF-8';
```
paulineribeyre: why do you need this?

suites/apis/centralizedAuth.js
```diff
@@ -0,0 +1,763 @@
...
+    fence.ask.assertStatusCode(
+        abc_read_response_after_force_create, 200,
+        msg='should have gotten authorized 200 for reading record under `/abc`'
+    );
+
+    // update
+    const filename_change = {
+      'file_name': 'test_filename'
+    }
+    const gen3_update_response = indexd.do.updateFile(
```
paulineribeyre:
```suggestion
    const gen3_update_response = await indexd.do.updateFile(
```

suites/apis/centralizedAuth.js
```diff
@@ -0,0 +1,763 @@
...
+        new_gen3_records.fooBarFile, filename_change, users.user2.accessTokenHeader)
+    const abc_update_response = indexd.do.updateFile(
```
paulineribeyre:
```suggestion
    const abc_update_response = await indexd.do.updateFile(
```

suites/apis/centralizedAuth.js
```diff
@@ -0,0 +1,763 @@
...
+      gen3_update_response,
+      'should NOT have got successful response when updating record under `/gen3`'
+    ).to.not.have.property('did');
+    chai.expect(
+      abc_update_response,
+      'should NOT have got successful response when updating record under `/abc`'
+    ).to.not.have.property('did');
+
+    // delete
+    const gen3_delete_response = indexd.do.deleteFile(
```
paulineribeyre:
```suggestion
    const gen3_delete_response = await indexd.do.deleteFile(
```

suites/apis/centralizedAuth.js
```diff
@@ -0,0 +1,763 @@
...
+        new_gen3_records.deleteMe, users.user2.accessTokenHeader)
+    const abc_delete_response = indexd.do.deleteFile(
```
paulineribeyre:
```suggestion
    const abc_delete_response = await indexd.do.deleteFile(
```

services/apis/indexd/indexdTasks.js
```diff
@@ -77,32 +81,85 @@ module.exports = {
     return true;  // always return true till we figure out the Promise.all issue above ...
   },
 
+  /**
+   * Fetches indexd res data for file and assigns 'rev', given an indexd file object with a did
+   * @param {Object} fileNode - Assumed to have a did property
+   * @returns {Promise<Gen3Response>}
```
paulineribeyre: update the docstring

services/apis/indexd/indexdTasks.js
```diff
@@ -77,32 +81,85 @@ module.exports = {
...
     ).then((res) => {
       file.rev = getRevFromResponse(res);
       return res.body;
     });
   },
 
+  /**
+   * Updates indexd data for file
+   * @param {Object} fileNode - Assumed to have a did property
+   * @returns {Promise<Gen3Response>}
```
paulineribeyre: also update this docstring... actually lots of docstrings in this file are not up to date :(

suites/apis/centralizedAuth.js
```diff
@@ -429,8 +434,8 @@ Scenario('User with access can CRUD indexd records in namespace, not outside nam
       abc_delete_response, new_abc_records.deleteMe,
       msg='user with permission could not delete record under `/abc`.'
     );
-    const getRes = await indexd.do.getFile(new_abc_records.deleteMe);
-    indexd.ask.recordDoesNotExist(getRes, new_abc_records.deleteMe);
+    const getRes = await indexd.do.getFileFullRes(new_abc_records.deleteMe);
+    indexd.ask.recordDoesNotExist(getRes);
```
paulineribeyre: `getFileFullRes ` returns the raw response but `recordDoesNotExist` takes a `Gen3Response` as param and looks for a `statusCode` prop

suites/apis/centralizedAuth.js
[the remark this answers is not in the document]
paulineribeyre: You haven't changed this, did you figure out what's up with the headers?

services/apis/fence/fenceTasks.js
Avantol13: good point, okay, will change it to a sequence

services/apis/indexd/indexdTasks.js
Avantol13: indexd dies without it :man_shrugging:

suites/apis/centralizedAuth.js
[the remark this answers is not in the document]
Avantol13: various indexd updates to support this now, so this should work passing through user tokens as expected

suites/apis/centralizedAuth.js
Avantol13: :+1:

services/apis/indexd/indexdTasks.js
Avantol13: :sob: will update them

suites/apis/centralizedAuth.js
Avantol13: the `statusCode` property should exist on regular responses too. but I can cast to this Gen3 class

suites/apis/centralizedAuth.js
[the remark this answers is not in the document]
paulineribeyre: oh, cool! then can you change the default parameter in `addFileIndices`, `updateFile` and `deleteFile` to be `accessTokenHeader` instead of `indexdAuthHeader`?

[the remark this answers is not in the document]
Avantol13: I think it would make other tests fail. `cdis.autotest@gmail.com` only has admin privileges in `abc` namespace for testing. So any other test trying to do other things would fail. I support I could make an additional admin user, but I figured we could wait to do that until we truly deprecate the basic auth that the current header is using

services/apis/indexd/indexdQuestions.js
```diff
@@ -38,9 +38,33 @@ module.exports = {
    * Asserts a record was successfully deleted from indexd
    * @param {Object} ffileNode
```
paulineribeyre:
```suggestion
   * @param {Object} fileNode
```
I may be looking at docstrings too much 😆 but can you add parameter `res` here (mostly to make it clear that it's the entire response)

services/apis/indexd/indexdQuestions.js
```diff
@@ -38,9 +38,33 @@ module.exports = {
...
+
+    if (res) {
+      const err = msg + 'Did not have expected status code of 200.';
+      expect(res, err).to.have.property('statusCode', 200);
+    }
+  },
+
+  /**
+   * Asserts a record was successfully deleted from indexd
+   * @param {Object} ffileNode
```
paulineribeyre:
```suggestion
   * @param {Object} fileNode
```
same here

services/apis/indexd/indexdTasks.js
```diff
@@ -133,9 +203,11 @@ module.exports = {
       var file = {
         did: guid
       };
-      await this.getFile(file); // adds 'rev' to the file
-      var res = await this.deleteFile(file);
-      fileList.push(file)
+      var fileRes = await this.getFile(file); // adds 'rev' to the file
+      if (!fileRes.error) {
```
paulineribeyre: could you add an `else` block to log that this file doesn't exist

services/apis/fence/fenceTasks.js
paulineribeyre: Copy of our discussion about this for future ref:
> I'm running into issues with it as a sequence, b/c hasUrl uses assertions, but before (for consistency in the tests) I was just getting the response (even if it was failed) and then doing an assert in the test (which I find clearer)
> if I keep it as a sequence, I can only call it when I expect a signedURL and need to check the responses where I expect to fail with essentially the check that was in the existing task
> tldr; so the main difference b/t just using the sequence and old way is that the old way was not doing asserts in the task b/c it's expected to fail for some cases
> it is returning an error message instead of throwing

PR #116, opened 2019-04-09
Feat/export to workspace test

mfshao:
Added tests for `export to workspace` feature, including tests for GUI (portal and JupyterHub) and tests for the manifest service API.

The test plan is specified at here: https://ctds-planx.atlassian.net/wiki/spaces/PLA/pages/42696711/Workspace+Automatic+Integration+Test+Coverage+Plan

Test scenarios including:
- Portal and JupyterHub/Hatchery:
1. Click Workspace tab when logged out and logged in
2. Check export to workspace message toaster
3. Export default manifest, mount it and check manifest name

- Manifest Service API:
1. Export default manifest and check if it exists in manifestservice endpoint

Verified in my dev env that all tests are passing.


services/portal/exportToWorkspace/exportToWorkspaceTasks.js
```diff
@@ -0,0 +1,113 @@
...
+      await this.checkAndAcceptJupyterDialogIfVisible();
+    });
+  },
+
+  /* Check if there is a modal dialog from Jupyther, if yes, accept it */
+  async checkAndAcceptJupyterDialogIfVisible() {
+    const ct = await I.grabNumberOfVisibleElements(exportToWorkspaceProps.modalDialogSelector);
+    if (ct === 1) {
+      const bt = await I.grabNumberOfVisibleElements(exportToWorkspaceProps.deleteConfirmButtonXPath);
+      if (bt === 1) {
```
paulineribeyre: same here, what happens if not 1?

services/portal/exportToWorkspace/exportToWorkspaceTasks.js
```diff
@@ -0,0 +1,113 @@
...
+  },
+
+  /* Spawn up a workspace */
+  startWorkspace() {
+    within({ frame: exportToWorkspaceProps.workspaceIFrameClass }, async () => {
+      const start = await I.grabNumberOfVisibleElements(exportToWorkspaceProps.startButtonSelector);
+      if (start === 1) { // see "Start My Server" button
+        I.click(exportToWorkspaceProps.startButtonSelector);
+        const spawn = await I.grabNumberOfVisibleElements(exportToWorkspaceProps.spawnButtonSelector);
+        if (spawn === 1) {
```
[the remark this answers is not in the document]
mfshao: actually both `start` and `spawn` can only be either 0 or 1. And if either of them is 0, means the  `mainWorkspaceApp` has already been spawned. So just wait for `mainWorkspaceApp` to become visible.

services/portal/exportToWorkspace/exportToWorkspaceTasks.js
mfshao: same here, both `ct` and `bt` can only be 0 or 1. If `ct` is 0 then no dialog is showing up, hence I think it should do nothing.

suites/portal/exportToWorkspaceTest.js
```diff
@@ -0,0 +1,31 @@
...
+Scenario('Click Workspace tab when logged out and logged in @exportToWorkspace', async (portalExportToWorkspace, home) => {
+  await portalExportToWorkspace.complete.checkClickWorkspaceTabWithLogoutAndLogin(home);
+});
+
+/* When clicking on the 'Export to Workspace' button, a message is displayed on the bottom of the screen. 
+On the successful request, the file name is specified in the message section. 
+If the request is not successful, there is an error message displaying the error that has occurred.
+Verify that the message toaster has correct format. */
+Scenario('Check export to workspace message toaster @exportToWorkspace', async (portalExportToWorkspace) => {
+  await portalExportToWorkspace.complete.checkMessageToaster();
```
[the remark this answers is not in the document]
mfshao: for now if `manifestservice`, `wts` and `portal` are all running right I think this will only gives you success toaster

services/portal/exportToWorkspace/exportToWorkspaceSequences.js
```diff
@@ -0,0 +1,56 @@
...
+  },
+
+  /* The 'Check export to workspace message toaster' test sequence, steps are:
+  - from 'Explore' page, click 'Export to workspace' button to export default manifest to workspace
+  - validate the format of toaster based on whether the export has succeeded or not */
+  checkMessageToaster() {
+    exportToWorkspaceTasks.exportDefaultManifestToWorkspace();
+    if (exportToWorkspaceQuestions.isManifestSavedToWorkspaceSucceeded()) {
+      exportToWorkspaceQuestions.doesSucceededMessageToasterLookCorrect();
+    } else if (exportToWorkspaceQuestions.isManifestSaveToWorkspaceFailed()) {
```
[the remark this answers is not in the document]
mfshao: right... I'll change this part

services/portal/exportToWorkspace/exportToWorkspaceTasks.js
```diff
@@ -0,0 +1,113 @@
...
+    const output = await within({ frame: exportToWorkspaceProps.workspaceIFrameClass }, () => {
+      I.waitForVisible(exportToWorkspaceProps.mainWorkspaceAppSelector, 600);
+      I.click(exportToWorkspaceProps.newDropDownButtonSelector);
+      I.click(exportToWorkspaceProps.newPython3ButtonSelector);
+      I.waitForVisible(exportToWorkspaceProps.pythonNotebookPanelSelector, 60); // wait for new notebook to show up
+      exportToWorkspaceProps.mountManifestCode.forEach((line) => {
+        I.pressKey(line);
+        I.pressKey('Enter');
+      });
+      I.wait(3); // wait for save button becomes clickable
```
paulineribeyre: you do `I.wait` several times, you could use the smartWait function to save a few seconds 😈

services/portal/exportToWorkspace/exportToWorkspaceSequences.js
```diff
@@ -0,0 +1,56 @@
...
+    await exportToWorkspaceTasks.deleteAllJupyterNotebooks();
+  },
+
+  /* The 'Click Workspace tab when logged out and logged in' test sequence, steps are:
+    - Log out current user
+    - Click the 'Workspace' tab from top nav bar to try to get to workspace
+    - Check if still seeing login page
+    - Login as test user
+    - Click the 'Workspace' tab from top nav bar again and verify is on workspace page */
+  checkClickWorkspaceTabWithLogoutAndLogin(home) {
```
paulineribeyre: these steps could be in the test itself instead of in a sequence, and you could turn this comment into a series of logs (one log before each step). the problem with functions like `seeElement` is that when they fail, the error message does not make it clear which feature failed. so the logs would let us see the progress and understand what failed

services/portal/exportToWorkspace/exportToWorkspaceSequences.js
```diff
@@ -0,0 +1,56 @@
...
+ */
+module.exports = {
+  /* The 'Export default manifest, mount it and check manifest name' test sequence, steps are:
+    - from 'Explore' page, click 'Export to workspace' button to export default manifest to workspace
+    - validate the message toaster and grab the filename of exported manifest file
+    - redirect to workspace and spawn a workspace if none exists
+    - create a python notebook in workspace and mount the latest manifest using python code
+    - verify the mounted file has the same filename as the exported manifest file
+    - go back to workspace and delete all Jupyther notebooks for clean up */
+  async checkExportDefaultManifestToWorkspace() {
```
paulineribeyre: same comment here

suites/portal/exportToWorkspaceTest.js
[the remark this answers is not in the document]
paulineribeyre: is it possible to test the failure toaster?

services/portal/exportToWorkspace/exportToWorkspaceTasks.js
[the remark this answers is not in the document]
paulineribeyre: in the tests, it should not have already been spawned, right? so they should always be 1?

services/portal/exportToWorkspace/exportToWorkspaceTasks.js
paulineribeyre: in which case would the dialog not be visible? i think maybe this should either be 2 separate functions, or this function should have an expected behavior

mfshao: In this case the dialog should always be visible. I was trying to make this method to be more generalize so if later we have a sequence that may not always have a visible dialog and we need to check before accepting we can use this as well.

services/portal/exportToWorkspace/exportToWorkspaceTasks.js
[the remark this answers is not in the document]
mfshao: Uh no. For a user on a common we only need to spawn a workspace once right? After spawning the second time we don't need to to this again. So if we don't check at here the tests will fail if the workspace has already been spawned

suites/portal/exportToWorkspaceTest.js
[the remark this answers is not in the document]
mfshao: I was hoping I can get a failure toaster by modifying user permissions but so far i got no luck. I'll try to figure it out.

services/portal/exportToWorkspace/exportToWorkspaceTasks.js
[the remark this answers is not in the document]
paulineribeyre: oh, i see
could you add a comment to the function to explain that?

services/portal/exportToWorkspace/exportToWorkspaceTasks.js
paulineribeyre: mm i see what you mean but in my opinion this function should just be `acceptJupyterDialog()` and you can do `I.see` and `I.click`, unless you have an idea of what use case would have a dialog that is sometimes visible, sometimes not

mfshao: agreed! I'll change this logic

services/portal/exportToWorkspace/exportToWorkspaceTasks.js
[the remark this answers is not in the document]
mfshao: sure

services/portal/exportToWorkspace/exportToWorkspaceSequences.js
mfshao: okay by logs you mean just `console.log`?

services/portal/exportToWorkspace/exportToWorkspaceTasks.js
mfshao: actually I'll remove this task and merge the `see` and `click` delete confirmation dialogic into `deleteAllJupyterNotebooks()`

services/portal/exportToWorkspace/exportToWorkspaceTasks.js
mfshao: I was not using `I.wait` here in the first place and hoping smartWait can take care of everything. But sometimes the test will fail at here so I have to add this arbitrary wait to let it work... Not sure why though 🤔

paulineribeyre: i meant this function https://github.com/uc-cdis/gen3-qa/blob/master/utils/apiUtil.js#L170
i see you use `I.wait(3)` several times and it probably doesn't take that long so you could use the smartwait

services/portal/exportToWorkspace/exportToWorkspaceSequences.js
paulineribeyre: yeah!

services/portal/exportToWorkspace/exportToWorkspaceSequences.js
```diff
@@ -1,56 +1,44 @@
...
     exportToWorkspaceTasks.exportDefaultManifestToWorkspace();
     exportToWorkspaceQuestions.isManifestSavedToWorkspaceSucceeded();
     const manifestFilename = await exportToWorkspaceTasks.grabManifestFilename();
     exportToWorkspaceTasks.jumpToWorkspacePage();
     exportToWorkspaceTasks.startWorkspace();
-    const mountOutput = await exportToWorkspaceTasks.mountLatestManifestInJupyterNotebook();
+    const mountOutput = await exportToWorkspaceTasks.mountLatestManifestInJupyterNotebook(exportToWorkspaceUtil);
     await exportToWorkspaceQuestions.doesMountOutputLookSuccessful(mountOutput, manifestFilename);
-    await exportToWorkspaceTasks.backToWorkspace();
+    await exportToWorkspaceTasks.backToWorkspace(exportToWorkspaceUtil);
```
paulineribeyre: `mountLatestManifestInJupyterNotebook` and `backToWorkspace` don't take a parameter

suites/portal/exportToWorkspaceTest.js
```diff
@@ -0,0 +1,30 @@
...
+On the successful request, the file name is specified in the message section. 
+Verify that the message toaster has correct format. */
+Scenario('Check export to workspace success message toaster @exportToWorkspace', async (portalExportToWorkspace) => {
+  await portalExportToWorkspace.complete.checkMessageToasterSuccess();
+});
+
+/* Export default manifest to workspace, mount the exported manifest using Python commands,
+and verify the mounted manifest file name should be same as the previously exported manifest.  */
+Scenario('Export default manifest, mount it and check manifest name @exportToWorkspace', async (portalExportToWorkspace, exportToWorkspaceUtil) => {
+  await portalExportToWorkspace.complete.checkExportDefaultManifestToWorkspace(exportToWorkspaceUtil);
```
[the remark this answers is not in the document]
mfshao: yep. accidentally put in some ongoing works... sorry

mfshao:
Was trying to make `exportToWorkspaceTasks.startWorkspace()` more stable. 
Found out previously this method will timeout after clicking `Spawn` button in random manners. Has tried several ways to stabilize it (`async/await`, `I.pressKey('Enter')`, `try/catch`, etc...).

The problem is when a workspace is spawning the codeceptjs seems cannot get the updated page automatically and it will timeouts. So far my solution is a bit hacky (it has multiple arbitrary `I.wait()`), but it can guarantee me the test can grabs the updates workspace app page each time after spawning a new one.

I'd appreciate any ideas that can helps me to make this method less hacky 😅

helpers/cdisHelper.js
```diff
@@ -17,6 +17,15 @@ class CDISHelper extends Helper {
       testResult.err.stack += res;
     });
   }
+
+  noTimeoutEnter() {
```
paulineribeyre: this function should probably be in the API utils file

mfshao: I feel like this function should be in a Helper file since it is customizing the `WebDriverIO helper`. Although it might be good if I move it to a separate `exportToWorkspaceHelper` file?

paulineribeyre: right now the API utils file contains anything that doesn't fit in another utils 😆but it's fine if you want to keep this function here

PR #117, opened 2019-04-19
Feat/no lock

paulineribeyre:
PXP-2842:
- remove locks
- use separate google projects for each jenkins env. for local tests, keep using the project that is currently being used (Gen3QA-ValidationJobTest)
- at the beginning of the tests, set up `googleProjectDynamic` depending on the namespace. make sure the fence monitor has access to the google project and delete existing sa keys on the "user service account". **Limitation:** codeceptjs doesn't let us send requests before the tests start, so we can clean in google but not in the fence db. this should be fine for jenkins since the db is reset before each run
- documentation

Bug fixes:
- fix pidgin backend test: test will not crash when there is a pidgin error
- do not run google tests in `nci-crdc-demo.datacommons.io` cdis-manifest PRs
- increase the login timeout (10sec->30sec) to avoid random login test failures
- errors during the bucket setup for the google tests will log a warning instead of throwing if tests are run locally
- `getUserGoogleCreds()` returned undefined if the GET response had no body. return `[]` instead and add logs for future debugging
- `createTempGoogleCreds()` similar issue: add logs for future debugging
- during SA registration, retry if there is a "socket hang up" error


docs/google-data-access-tests.md
```diff
@@ -0,0 +1,44 @@
...
+| planxparentproject       | dummy-one@planx-pla.net | googleProjectWithParentOrg          | Used to test invalid SA registration |
+| projectfencenoaccess     | gen3.autotest@gmail.com | googleProjectFenceNotRegistered     | Used to test invalid SA registration |
+| ProjectServiceAcctHasKey | gen3.autotest@gmail.com | googleProjectServiceAcctHasKey      | Used to test invalid SA registration |
+| Gen3QA-[namespace]       | gen3.autotest@gmail.com | googleProjectDynamic                | See "Notes on googleProjectDynamic" section |
+
+| Service account                                                 | Purpose |
+|-----------------------------------------------------------------|---------|
+| fence-service@dcf-integration.iam.gserviceaccount.com           | Fence monitor. Needs "Editor" and "Project IAM Admin" permissions on the Google projects |
+| gen3qa-service@gen3qa-validationjobtest.iam.gserviceaccount.com | Service account used by the tests to interact with the Google projects. It has access to everything |
+| service-account@gen3qa-[namespace].iam.gserviceaccount.com      | Represents a user who registers a Google project. Each googleProjectDynamic has one |
```
Avantol13:
```suggestion
| service-account@gen3qa-[namespace].iam.gserviceaccount.com      | Represents a researcher's project SA for "registering" for controlled access data through Gen3. Each googleProjectDynamic has one |
```

docs/google-data-access-tests.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Notes on googleProjectDynamic
+
+Most Google projects that have been set up for testing purposes are not modified by the tests.  However, some tests require modifying the Google project itself (service account external key creation/deletion, permission updates...). If several of these tests are running simultaneously, they will make each other fail - for example, if a test generates an external key for a service account, other tests will not be able to register service accounts in the same project.
```
Avantol13:
```suggestion
Most Google projects that have been set up for testing purposes are not modified by the tests.  However, some tests require modifying the Google project itself (service account external key creation/deletion, permission updates...). If several of these tests are running simultaneously, they will make each other fail - for example, if a test generates an external key for a service account, other tests will not be able to register service accounts in the same Google project.
```

docs/google-data-access-tests.md
```diff
@@ -0,0 +1,44 @@
...
+
+All the tests that require exclusive access to a Google Project use the same project: `googleProjectDynamic`. Each Jenkins environment has its own "dynamic project" that it has exclusive access to. During the test setup, we make sure to reset this project to what it should be before the tests modify it.
+
+When setting up a new Jenkins environment, we need to set up a new Google project:
```
Avantol13: are there docs elsewhere on how to set up a new jenkins env? might want to link to this there if it exists

services/apis/fence/fenceProps.js
```diff
@@ -34,7 +34,6 @@ module.exports = {
     registerGoogleServiceAccount: `${rootEndpoint}/google/service_accounts`,
     deleteGoogleServiceAccount: `${rootEndpoint}/google/service_accounts`,
     getGoogleServiceAccounts: `${rootEndpoint}/google/service_accounts`,
-    getGoogleServiceAccountMonitor: `${rootEndpoint}/google/service_accounts/monitor`,
```
Avantol13: is this change necessary? we should still be able to use this for the monitoring SA

Avantol13: I would rather not hard-code the SA if possible

services/apis/fence/fenceProps.js
```diff
@@ -351,14 +352,16 @@ module.exports = {
     owner: 'gen3.autotest@gmail.com',
   },
 
-  // used when the tests need to modify the google project/SAs
+  // used when the tests need to modify the google project itself
+  // the id is updated during test_setup depending on the current namespace
```
Avantol13:
```suggestion
  // NOTE: the id is updated during test_setup depending on the current namespace
```

utils/google.js
```diff
@@ -406,127 +426,4 @@ module.exports = {
       });
     })
   },
-
```
Avantol13: it's kinda sad to see this all deleted :cry:

Avantol13: it's better this way. Rest in peace locking code

test_setup.js
```diff
@@ -151,24 +152,87 @@ function assertGen3Client() {
...
+  ];
+  for (var role of monitorRoles) {
+    let res = await google.updateUserRole(
+      fenceProps.googleProjectDynamic.id,
+      {
+        role,
+        members: [`serviceAccount:${fenceProps.monitorServiceAccount}`]
+      }
+    );
+    if (res.code) {
```
Avantol13: shouldn't this check be something like `if res.code != 200`?

services/apis/fence/fenceProps.js
paulineribeyre: codeceptjs doesn't let us send requests during the test setup :(

utils/google.js
paulineribeyre: 💔

test_setup.js
paulineribeyre: if it works there's no `code` at all 🤷‍♀️

Avantol13: :confounded: welp, makes sense then

PR #118, opened 2019-05-13
Feat/multipart upload

paulineribeyre:
Add test for multipart upload endpoints in fence


services/apis/fence/fenceSequences.js
```diff
@@ -106,6 +106,52 @@ module.exports = {
...
+  /**
+   * Hits fence's multipart upload initialization endpoint
+   * @param {string} fileName - name of the file that will be uploaded
+   * @param {string} accessToken - access token
+   * @returns {object} { guid, uploadId }
+   */
+  async initMultipartUpload(fileName, accessHeader) {
+      const res = await fenceTasks.initMultipartUpload(fileName, accessHeader);
+      expect(res, 'Unable to initialize multipart upload').to.have.nested.property('body.guid');
+      expect(res, 'Unable to initialize multipart upload').to.have.nested.property('body.uploadId');
```
giangbui: Should check `status_code` == 201 here?

suites/apis/dataUploadTest.js
```diff
@@ -266,7 +267,75 @@ Scenario('Upload the same file twice @dataUpload', async (sheepdog, indexd, node
...
+      file_size: files.getFileSize(bigFileName)
+    }
+  };
+  await dataUpload.waitUploadFileUpdatedFromIndexdListener(indexd, fileNode);
+
+  // download the file
+  console.log("Downloading the file");
+  const signedUrlRes = await fence.do.createSignedUrlForUser(fileGuid);
+  fence.ask.assertStatusCode(signedUrlRes, 200, "Could not get signed URL for file download");
+  fence.complete.checkFileEquals(signedUrlRes, fileContents.toString());
```
giangbui: I prefer to check hash and size than than check the entire object content since it follows our logic but it is OK for me if you want to keep that line

suites/apis/dataUploadTest.js
```diff
@@ -266,7 +267,75 @@ Scenario('Upload the same file twice @dataUpload', async (sheepdog, indexd, node
   );
 }).retry(2);
 
-BeforeSuite(async (dataClient, fence, users, sheepdog, indexd, files) => {
+/**
+ * Use fence's multipart upload endpoints to upload a large data file (>5MB)
+ */
+Scenario('Multipart upload', async (files, users, fence, indexd, dataUpload) => {
```
giangbui: It would be great if you have some tests for failure cases such as wrong etag when completion?

suites/apis/dataUploadTest.js
paulineribeyre: the hash and size are checked as well, during `waitUploadFileUpdatedFromIndexdListener`

PR #121, opened 2019-05-23
feat(run-tests): attempt to get version to make decisions on what tes…

Avantol13:
…ts to run

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -82,18 +82,6 @@ runServiceTestsIfVersion() {
   fi
 }
 
-
-# environments that use DCF features
-# we only run Google Data Access tests for cdis-manifest PRs to these
-envsRequireGoogle="dcp.bionimbus.org gen3.datastage.io nci-crdc-staging.datacommons.io nci-crdc.datacommons.io"
-
-
-#
-# DataClientCLI tests require a fix to avoid parallel test runs
-# contending over config files in the home directory
-#      
-doNotRunRegex="@dataClientCLI"
-
 # little helper to maintain doNotRunRegex
 donot() {
   local or
@@ -106,9 +94,6 @@ donot() {
   fi
   doNotRunRegex="${doNotRunRegex}${or}$1"
 }
-
-# Do not run performance testing
-donot '@Performance'
 
 #----------------------------------------------------
 # main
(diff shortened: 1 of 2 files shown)


run-tests.sh
```diff
@@ -31,6 +31,57 @@ dryrun() {
...
+  # Get last item of delimited string using string operators:
+  #    https://www.linuxjournal.com/article/8919
+  version=${response##*:}
+  echo $version
+}
+
+# Takes 3 arguments:
+#   $1 service name
+#   $2 test tag to avoid until version is greater than next arg
+#   $3 versio of service where tests apply >=
```
paulineribeyre:
```suggestion
#   $3 version of service where tests apply >=
```

run-tests.sh
```diff
@@ -31,6 +31,57 @@ dryrun() {
...
+#
+# ex: runServiceTestsIfVersion "fence" "@multipartupload" "3.0.0"
```
paulineribeyre: not important, but the order of the arguments bothers me a little bit haha
`runServiceTestsIfVersion "@multipartupload" "fence" "3.0.0"` would look more natural to me

run-tests.sh
```diff
@@ -31,6 +31,57 @@ dryrun() {
...
+runServiceTestsIfVersion() {
+  local currentVersion
+  currentVersion=$(getServiceVersion $1)
```
paulineribeyre: idk much about bash style conventions but should you make sure the arguments are provided?

run-tests.sh
Avantol13: I did think it seems a little weird, I'll change it to your suggestion

run-tests.sh
```diff
@@ -135,8 +176,23 @@ EOM
...
 
+# environments that use DCF features
+# we only run Google Data Access tests for cdis-manifest PRs to these
+envsRequireGoogle="dcp.bionimbus.org gen3.datastage.io nci-crdc-staging.datacommons.io nci-crdc.datacommons.io"
+
+#
+# DataClientCLI tests require a fix to avoid parallel test runs
+# contending over config files in the home directory
+#
+doNotRunRegex="@dataClientCLI"
```
paulineribeyre:
```suggestion
donot '@dataClientCLI'
```
`doNotRunRegex` should be initialized earlier

run-tests.sh
```diff
@@ -56,8 +100,7 @@ donot() {
   doNotRunRegex="${doNotRunRegex}${or}$1"
```
paulineribeyre: mm i was wrong :p `doNotRunRegex` does not need to be initialized

paulineribeyre:
a cool improvement would be to allow multiple service versions: for rbac you need latest fence, indexd and arborist

PR #122, opened 2019-05-28
feat(central-auth): add tests for public files and consent codes

Avantol13:


suites/apis/centralizedAuth.js
```diff
@@ -154,19 +154,48 @@ const indexed_files = {
...
+    filename: 'testdata',
+    link: 's3://cdis-presigned-url-test/testdata',
+    md5: '73d643ec3f4beb90213ef0beed440af1',
+    authz: [
+      '/abc/programs/test_program/projects/test_project1',
+      '/consents/HMB'
+    ],
+    size: 44,
+  },
+  gruResearchFile: {
```
paulineribeyre: i think this file is the same as `generalResearchUseFile`? and `generalResearchUseFile` is unused. could you remove one of the 2?

suites/apis/centralizedAuth.js
```diff
@@ -736,28 +765,131 @@ Scenario('Client with user token WITH permission CAN create signed URL for recor
...
+});
 
 
 /*
    - Give user access to program/proj and HMB policy, make sure they CAN access
      files with GRU
-*/
\ No newline at end of file
+*/
+Scenario('Test create signed URL for file in authorized namespace with IMPLIED authorized consent code (based on DUO hierarchy) @centralizedAuth',
```
paulineribeyre: i don't understand the point of this test, to me it looks like it tests the same thing as the first test (`Test create signed URL for file in authorized namespace with authorized consent code (multiple policies`)

`HMB ` and `GRU` are both in policy `hmb-researcher` so both tests check that the user has access to "any" code that's in the policy

can you explain? :D

suites/apis/centralizedAuth.js
Avantol13: oh, yeah, whoops :man_facepalming:

suites/apis/centralizedAuth.js
Avantol13: yeeah, so the way this is going to appear to work to end-users is you give a user a single policy for what they've consented to and there is an _implied_ hierarchy for the codes. I'll send you links to some diagrams :confetti_ball: on slack.

basically, by having access to HMB, user should also have access to GRU. so this functionality may be somewhat duplicated in other tests that have multiple policies, but this is a specific test case for the consent codes themselves.

I guess it's all configuration, so maybe we don't really _need_ this test :man_shrugging: but it tests the client user case explicitly better than the first test.

I leaned towards writing it to be more explicit about the expectation of the consent codes but I guess I could be convinced to remove it

paulineribeyre: ✔️

PR #136, opened 2019-06-12
feat(dbgap): dbgap tests and supporting functionality

Avantol13:


test_setup.js
```diff
@@ -370,12 +382,12 @@ module.exports = async function (done) {
...
     }
 
-    if (isIncluded('@reqGoogle')) {
+    if (isIncluded('@reqGoogle') || isIncluded('@dbgapSyncing')) {
       createGoogleTestBuckets();
       await setupGoogleProjectDynamic();
     }
 
-    // Create a program and project (does nothing if already exists)
+    Create a program and project (does nothing if already exists)
```
paulineribeyre:
```suggestion
    // Create a program and project (does nothing if already exists)
```
:p

suites/apis/dbgapTest.js
```diff
@@ -0,0 +1,280 @@
...
+
+    // asserts for update
+    chai.expect(
+      dbgap_update_response,
+      'should NOT have got successful response when updating record under `/dbgap`'
+    ).to.not.have.property('did');
+
+    // delete
+    const dbgap_delete_response = await indexd.do.deleteFile(
+        new_dbgap_records.deleteMe, users.user2.accessTokenHeader
```
paulineribeyre: can you just delete `fooBarFile` instead of a different file?

suites/apis/dbgapTest.js
```diff
@@ -0,0 +1,280 @@
...
+
+    new_dbgap_records.fooBarFile.did = dbgapFooBarFileGuid;
+    new_dbgap_records.deleteMe.did = dbgapDeleteMe;
+
+    // create
+    const dbgap_create_success = await indexd.do.addFileIndices(
+        Object.values(new_dbgap_records), users.user2.accessTokenHeader
+    );
+
+    // read to test creation
```
paulineribeyre: the tests in the `centralizedAuth` test suite check record creation the same way, and i guess trying to read the record can be a good extra check, but you can start with:
```
const ok = await indexd.do.addFileIndices();
chai.expect(ok).to.be.true;
```

test_setup.js
Avantol13: :man_facepalming:

suites/apis/dbgapTest.js
```diff
@@ -0,0 +1,280 @@
...
+      `User ${users.mainAcct.username} with access could NOT create s3 signed urls ` +
+      'and read file for a record in authorized dbGaP project phs000178'
+    ).to.equal(fence.props.awsBucketInfo.cdis_presigned_url_test.testdata);
+    chai.expect(QAFileContents,
+      `User ${users.mainAcct.username} with access could NOT create s3 signed urls ` +
+      'and read file for a record in authorized program QA'
+    ).to.equal(fence.props.awsBucketInfo.cdis_presigned_url_test.testdata);
+});
+
+Scenario('dbGaP + user.yaml Sync (from prev test): ensure user cannot create/update/delete dbGaP indexd records @dbgapSyncing',
```
paulineribeyre:
```suggestion
Scenario('dbGaP + user.yaml Sync (from prev test): ensure user without dbGap access cannot create/update/delete dbGaP indexd records @dbgapSyncing',
```

paulineribeyre: maybe add a test `ensure user with dbGap access can create/update/delete dbGaP indexd records`

suites/apis/dbgapTest.js
paulineribeyre: `addFileIndices` creates 2 records and you only test reading one
you could just add one of them (i just noticed we don't even have an indexd function to add a singe record haha)

utils/bash.js
```diff
@@ -22,11 +22,14 @@ class Bash{
       return new Docker();
   }
 
-  runJob(jobName, args='') {
-    return this.commander.runJobAndWait(jobName, args);
+  runJob(jobName, args='', wait=true) {
```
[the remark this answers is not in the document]
Avantol13: whoops, I switched back to waiting to make sure sync happens before other tests run. I had it on the suite cleanup but ended up not using it. It was working at the time, I just changed it back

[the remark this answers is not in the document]
Avantol13: Not sure if there are use cases for this down the line.. I could just remove it

suites/apis/dbgapTest.js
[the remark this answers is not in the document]
Avantol13: it does take a long time :cry: I have to do some more tests to check how long it actually takes, but it's definitely more than 20 seconds

suites/apis/dbgapTest.js
```diff
@@ -0,0 +1,280 @@
...
+      signedUrls3phs000178Res);
+    let phs000178gsFileContents = await fence.do.getFileFromSignedUrlRes(
+      signedUrlgsPhs000178Res);
+
+    let phst000179s3FileContents = await fence.do.getFileFromSignedUrlRes(
+      signedUrls3phs000179Res);
+    let phst000179gsFileContents = await fence.do.getFileFromSignedUrlRes(
+      signedUrlgsPhs000179Res);
+
+    chai.expect(phs000178s3FileContents,
```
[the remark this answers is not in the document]
Avantol13: ah, you're probably right :+1: force of habit

utils/bash.js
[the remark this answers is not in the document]
paulineribeyre: we can keep it as long as we know it works!

suites/apis/dbgapTest.js
```diff
@@ -0,0 +1,280 @@
...
+    ok, 'unable to add files to indexd as part of dbgapTest setup').to.be.true;
+});
+
+AfterSuite(async (fence, indexd, users) => {
+  console.log('Removing indexd files used to test signed urls');
+  await indexd.do.deleteFileIndices(Object.values(indexed_files));
+  await indexd.do.deleteFileIndices(Object.values(new_dbgap_records));
+
+  console.log('Running usersync after dbgap testing');
+  bash.runJob('usersync', args='FORCE true');
```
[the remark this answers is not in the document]
Avantol13: usersync does a diff between the "last synced user.yaml" and if it doesn't find any differences, it won't run (even if arborist needs to be updated). So for the tests I've been FORCE-ing it to ensure that it actually runs. I can update the other places too

suites/apis/dbgapTest.js
```diff
@@ -0,0 +1,280 @@
...
+    ).to.equal(fence.props.awsBucketInfo.cdis_presigned_url_test.testdata);
+    chai.expect(phs000178gsFileContents,
+      `User ${users.mainAcct.username} with access could NOT create gs signed urls ` +
+      'and read file for a record in authorized dbGaP project phs000178'
+    ).to.equal(fence.props.googleBucketInfo.test.fileContents);
+
+    chai.expect(phst000179s3FileContents,
+      `User ${users.mainAcct.username} WITHOUT access COULD create s3 signed urls ` +
+      'and read file for a record in unauthorized dbGaP project phs000179'
+    ).to.not.equal(fence.props.awsBucketInfo.cdis_presigned_url_test.testdata);
```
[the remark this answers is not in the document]
Avantol13: oh, I think I just felt the exporting of the global and syntax for getting it was weird. like I'd have to do `.to.equal(fence.do.FILE_FROM_URL_ERROR)`. which I guess is okay

suites/apis/dbgapTest.js
Avantol13: oh, this specific test is checking for a `404` to make sure it did _not_ get created. THe next section `force creation of records` does just check if the response is true

suites/apis/dbgapTest.js
Avantol13: well, so that's sorta the weird part. We actually don't want just anyone with access to be able to modify the indexd record. We'd have to give special access to data nodes who need to create records for a specific project, but anyone with access should only have read

suites/apis/dbgapTest.js
Avantol13: oh, yeah, that should be fine. I'll update

suites/apis/dbgapTest.js
paulineribeyre: then do you think a `ensure user with dbGap read access cannot create/update/delete dbGaP indexd records` test (using mainAcct) would be useful? maybe both or maybe in replacement of this one

suites/apis/dbgapTest.js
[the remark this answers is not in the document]
paulineribeyre: maybe you can make it a Prop?

suites/apis/dbgapTest.js
paulineribeyre: oh right 😆

suites/apis/dbgapTest.js
```diff
@@ -0,0 +1,270 @@
...
+      `User ${users.mainAcct.username} with access could NOT create s3 signed urls ` +
+      'and read file for a record in authorized dbGaP project phs000178'
+    ).to.equal(fence.props.awsBucketInfo.cdis_presigned_url_test.testdata);
+    chai.expect(QAFileContents,
+      `User ${users.mainAcct.username} with access could NOT create s3 signed urls ` +
+      'and read file for a record in authorized program QA'
+    ).to.equal(fence.props.awsBucketInfo.cdis_presigned_url_test.testdata);
+});
+
+Scenario('dbGaP + user.yaml Sync (from prev test): ensure user without dbGap access cannot create/update/delete dbGaP indexd records @dbgapSyncing',
```
paulineribeyre: does this one not need @reqGoogle?

suites/apis/dbgapTest.js
Avantol13: yeah, I could update this test to check user with _and_ without dbgap access, since in both cases they shouldn't be able to create/update/delete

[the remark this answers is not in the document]
Avantol13: It's worse than I thought :grimacing: , at least in my dev env. I'll leave the timestamps in the logs so we can check QA. Did some timing. Just dbgap sync job took `81` seconds, dbgap+user.yaml took `93` seconds

suites/apis/dbgapTest.js
```diff
@@ -93,14 +114,19 @@ AfterSuite(async (fence, indexd, users) => {
   await indexd.do.deleteFileIndices(Object.values(indexed_files));
   await indexd.do.deleteFileIndices(Object.values(new_dbgap_records));
 
-  console.log('Running usersync after dbgap testing');
-  bash.runJob('usersync', args='FORCE true');
+  // console.log('Running usersync after dbgap testing');
+
+  console.log('start time: ' + Math.floor(Date.now() / 1000))
+  // bash.runJob('usersync', args='FORCE true');
```
paulineribeyre: did you comment this out on purpose?

suites/apis/dbgapTest.js
```diff
@@ -0,0 +1,389 @@
...
+      `User ${users.mainAcct.username} with access could NOT create s3 signed urls ` +
+      'and read file for a record in authorized dbGaP project phs000178'
+    ).to.equal(fence.props.awsBucketInfo.cdis_presigned_url_test.testdata);
+    chai.expect(QAFileContents,
+      `User ${users.mainAcct.username} with access could NOT create s3 signed urls ` +
+      'and read file for a record in authorized program QA'
+    ).to.equal(fence.props.awsBucketInfo.cdis_presigned_url_test.testdata);
+});
+
+Scenario('dbGaP + user.yaml Sync (from prev test): ensure user without dbGap access cannot create/update/delete dbGaP indexd records @dbgapSyncing',
```
paulineribeyre: this test is missing @reqGoogle

suites/apis/dbgapTest.js
Avantol13: nope :man_facepalming:

suites/apis/dbgapTest.js
```diff
@@ -129,30 +155,52 @@ Scenario('dbGaP Sync: created signed urls (from s3 and gs) to download, try crea
...
       indexed_files.phs000179File.did, ['protocol=s3'],
       users.mainAcct.accessTokenHeader
     );
-    console.log('Use mainAcct to create gs signed URL for file phs000179')
+    console.log('Use mainAcct to create s3 signed URL for phs000179 file in orgB namespace')
+    const signedUrls3anotherPhs000179FileRes = await fence.do.createSignedUrl(
+      indexed_files.anotherPhs000179File.did, ['protocol=s3'],
+      users.mainAcct.accessTokenHeader
+    );
+    console.log('Use mainAcct to create gs signed URL for file phs000179 in default namespace')
```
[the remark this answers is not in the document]
Avantol13: I think I changed it so default is `/` but namespaces for `phs000179` are `/orgA/` and `/orgB`. I will update the logs to make more sense

services/apis/indexd/indexdTasks.js
```diff
@@ -40,12 +40,20 @@ module.exports = {
...
         hashes: { md5: file.md5 },
         acl: file.acl,
         metadata: file.metadata,
       };
 
+      if (file.urls) {
+        data.urls = file.urls
+      } else if (file.link) {
+        data.urls = [file.link];
+      } else {
```
paulineribeyre: you don't need this `else` since `urls` is initialized as []

PR #155, opened 2019-07-17
Fix/ccs

vpsx:
https://github.com/uc-cdis/sheepdog/pull/294 
^ The indexd record authz field will now get updated with consent codes during data upload flow. This presents a problem when consent codes that are randomly generated by the data simulator end up in the node, causing the test accounts to lose access. This PR changes submitGraphAndFileMetadata such that it will not use/submit the randomly generated CCs, and instead will take CCs as an argument. 

### Bug Fixes
get consent codes from fn arg instead of from node property generated by data simulator


paulineribeyre:
fixed by https://github.com/uc-cdis/data-simulator/pull/27

PR #168, opened 2019-09-06
fix(auth-error): handle both pre- and post-arborist

paulineribeyre:
Before sheepdog's integration with arborist, the error is "Authentication Error: Signature has expired" and it's in body.message
After, the error is "request to arborist failed: error decoding token: expired at time: 1567720295" and it's in body.error
This works for both cases

### DIFF
--- a/services/apis/sheepdog/sheepdogProps.js
+++ b/services/apis/sheepdog/sheepdogProps.js
@@ -62,12 +62,12 @@ module.exports = {
   /**
    * Expected contents of the response when no authentication is provided
    */
-  resNoAuth: {
+  resExpiredAuth: {
     status: 401,
     // Before sheepdog's arborist update, the error is:
     // "Authentication Error: Signature has expired"
     // After sheepdog's arborist update, the error is:
-    // "request to arborist failed: error decoding token: expired at time: 1567720295"
+    // "request to arborist failed: error decoding token: expired at time: 123456"
     errorMsg: 'expired',
   },
 
--- a/services/apis/sheepdog/sheepdogProps.js
+++ b/services/apis/sheepdog/sheepdogProps.js
@@ -60,13 +60,16 @@ module.exports = {
   }),
 
   /**
-   * Gen3Response when no authentication provided
+   * Expected contents of the response when no authentication is provided
    */
-  resNoAuth: new Gen3Response({
-    request: {},
+  resNoAuth: {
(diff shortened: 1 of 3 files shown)


services/apis/sheepdog/sheepdogProps.js
```diff
@@ -60,13 +60,16 @@ module.exports = {
   }),
 
   /**
-   * Gen3Response when no authentication provided
+   * Expected contents of the response when no authentication is provided
```
vpsx: 🙇

services/apis/sheepdog/sheepdogProps.js
```diff
@@ -60,13 +60,16 @@ module.exports = {
...
    */
-  resNoAuth: new Gen3Response({
-    request: {},
+  resNoAuth: {
```
vpsx: If this is a generic 401 missing auth, then maybe the message shouldn't be specifically about expiry? 
But it seems that the only place that this (and `hasNoAuthError`) are used is this one scenario in `SubmitAndQueryNodesTest`:
```
  5 Scenario('submit node unauthenticated @reqData', async (sheepdog, nodes, users) => {
  6   await sheepdog.do.addNode(nodes.getFirstNode(), users.mainAcct.expiredAccessTokenHeader);
  7   sheepdog.ask.hasNoAuthError(nodes.getFirstNode().addRes);
[..blah]
```
Since this is testing with `expiredAccessTokenHeader` and this is the only place `hasNoAuthError` is used, maybe just rename `hasNoAuthError` --> `hasExpiredAuthError` and `resNoAuth` --> `resExpiredAuth`? :pray:

services/apis/sheepdog/sheepdogProps.js
```diff
@@ -60,13 +60,16 @@ module.exports = {
...
     status: 401,
-    body: { message: 'Authentication Error: Signature has expired' },
-  }),
+    // Before sheepdog's arborist update, the error is:
+    // "Authentication Error: Signature has expired"
+    // After sheepdog's arborist update, the error is:
+    // "request to arborist failed: error decoding token: expired at time: 1567720295"
```
vpsx: Oh also fwiw this is a real timestamp 😆

paulineribeyre: yeah i thought it was fine but i can change it :p

PR #174, opened 2019-10-10
Fix/core metadata

mfshao:
Fixed core metadata page integration tests

Depends on https://github.com/uc-cdis/data-portal/pull/589


suites/portal/coremetadatapageTest.js
```diff
@@ -1,40 +1,30 @@
...
+Feature('CoreMetadataPageTest');
+
+const I = actor();
+
+BeforeSuite(async (nodes, sheepdog, pidgin, users) => {
+  await sheepdog.complete.findDeleteAllNodes();
+  await sheepdog.complete.addNodes(nodes.getPathToFile());
+  const validFile = nodes.getFileNode().clone();
+  await sheepdog.complete.addNode(validFile);
+  metadata = await pidgin.do.getCoremetadata(validFile, 'application/json', users.mainAcct.accessTokenHeader);
```
paulineribeyre: i know this line depends on the backend not the frontend, but i still think it should be in the test, not in BeforeSuite. and it should include `pidgin.ask.seeJsonCoremetadata(valid_file, metadata);` too

When we see the tests failed during "BeforeSuite" the initial reaction is to think it's not an actual failing test

services/portal/coreMetadataPage/coreMetadataPageProps.js
```diff
@@ -0,0 +1,24 @@
...
+  coreMetadataPageBox2Class: '.core-metadata-page__box2',
+  coreMetadataPageBox3Class: '.core-metadata-page__box3',
+  //
+  // ready_cue: {
+  //   locator: {
+  //     css: '.this-element-appears-when-page-loaded'
+  //   }
+  // },
+  //
+  // filterTab: {
```
paulineribeyre: could you remove the commented out stuff

suites/portal/coremetadatapageTest.js
```diff
@@ -1,40 +1,30 @@
...
+});
+
+Before((home) => {
+  home.complete.login();
+});
+
+Scenario('test core metadata page @coreMetadataPage @portal', async (portalCoreMetadataPage) => {
+  await portalCoreMetadataPage.complete.checkFileCoreMetadataPage(metadata);
```
paulineribeyre: another test that starts from the "files" tab of the explorer, clicks on the file and checks the core metadata page would be good
i don't know if it would be easy to do though since it now depends on the explorer page?
if you think we can't do it for now, could you still add a "todo" comment?

services/portal/coreMetadataPage/coreMetadataPageQuestions.js
```diff
@@ -0,0 +1,57 @@
...
+  return `${fileSizeStr} ${suffix}`;
+};
+
+const dateTransform = date => `Updated on ${date.substr(0, 10)}`;
+
+/**
+ * coreMetadataPage Questions
+ */
+module.exports = {
+  doesCoreMetadataPageLooksCorrect(metadata) {
```
paulineribeyre:
```suggestion
  doesCoreMetadataPageLookCorrect(metadata) {
```

services/portal/coreMetadataPage/coreMetadataPageQuestions.js
```diff
@@ -0,0 +1,57 @@
...
+
+    for (const metadataKey of Object.keys(metadata)) {
+      switch (metadataKey) {
+        // no show for 'project_id'
+        case 'project_id':
+          I.dontSee(metadata[metadataKey]);
+          break;
+        // special cases for 'file_size', 'type' and 'updated_datetime'
+        case 'file_size':
+          I.see(fileSizeTransform(metadata[metadataKey]));
```
paulineribeyre: i'm not sure copying the transform functions from data-portal is the way to go 🤔 
i would ignore project_id, type, file_size and updated_datetime (and keep the comments that explain why) and just look for other fields as well as "more data info" and "download" button
this is just a suggestion though :)

suites/portal/coremetadatapageTest.js
mfshao: yeah I was also thinking of this initally. But for now it is hard to do because the new file explorer is powered by guppy and es. And in integration tests now we are pinning the es indices for all envs to certain indices so it won't work unfortunately 😞

but I can definitely put a TODO in there

services/portal/coreMetadataPage/coreMetadataPageQuestions.js
mfshao: a ha! good call, I should check for the button and table title

services/portal/coreMetadataPage/coreMetadataPageQuestions.js
```diff
@@ -3,55 +3,36 @@ const coreMetadataPageProps = require('./coreMetadataPageProps.js');
...
-        case 'updated_datetime':
-          I.see(dateTransform(metadata[metadataKey]));
-          break;
-        // all other fields, showing normally
-        default:
-          I.see(metadata[metadataKey]);
-          break;
+    Object.keys(metadata).forEach(metadataKey => {
+      if (excludedFieldList.includes(metadataKey)) {
+        I.dontSee(metadata[metadataKey]);
```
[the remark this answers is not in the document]
mfshao: yeah you are absolutly right. Somehow I lost my mind while writing that code... 🤦‍♂

PR #175, opened 2019-10-15
test(dcfreplication): executable test plan added

atharvar28:
This is an executable test plan for dcf_dataservice data replication service. 

This PR included :
1. AWS sync
2. Google sync
3. check for Updated Indexd

### DIFF
--- a/suites/apis/dcfDataReplicationTest.js
+++ b/suites/apis/dcfDataReplicationTest.js
@@ -1,0 +1,86 @@
+Feature('DCF dataservice Data Replication - https://ctds-planx.atlassian.net/browse/PXP-3483');
+
+const chai = require('chai');
+const {interactive, ifInteractive} = require('../../utils/interactive.js');
+const expect = chai.expect;
+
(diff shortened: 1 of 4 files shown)


suites/apis/dcfDataReplicationTest.js
```diff
@@ -0,0 +1,86 @@
+Feature('DCF dataservice Data Replication - https://ctds-planx.atlassian.net/browse/PXP-3483');
+
+const chai = require('chai');
+const {interactive, ifInteractive} = require('../../utils/interactive.js');
+const expect = chai.expect;
+
+Scenario('AWS Sync - create a directory in $vpc_name/apis_configs called dcf_dataservice', ifInteractive(
```
frickjack: * Should we put a `@manual` annotation on these till they are automated?
* Add `donot @manual` to `run-tests.sh`
* Also go through the other manual tests and do the same thing.
* And rename `portal/dashboardReports.js` and `portal/roleBasedUI.js` to `portal/dashboardReportsTest.js` and `portal/`roleBasedUITest.js`

PR #178, opened 2019-10-25
Chore/token from env

frickjack:


getAccessToken.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+help() {
+  cat - <<EOM
+  Use: baseh ./getAccessToken.sh apiKey
```
themarcelor: Is that a typo?
`baseh` -> `bash`

README.md
```diff
@@ -11,28 +11,47 @@ gen3 integration tests - run by https://jenkins.planx-pla.net/ via a `Jenkinsfil
...
-```
-# skip program/project creation at startup for now ...
+The `getAccessToken.sh` script is a little helper that can retrieve an access token given an API key.
+
+* First - generate some test data (if necessary for the test) with the appropriate dictionary (see below)
+* Next - generate an access token (it will expire in 20 minutes) using an API key from the target commons
 
+```
 mkdir -p output
+export GEN3_TOKEN_MAIN=$(bash ./getAccessToken.sh ~/Secrets/gen3/qa-brain.planx-pla.net/credentials.json)
```
themarcelor: Should we consider some err check here? e.g.,
```
output=$(bash ./getAccessToken.sh ~/Secrets/gen3/qa-brain.planx-pla.net/credentials.json)
RC=$?
if [ $RC -ne 0 ]; then
  echo "something went wrong while trying to obtain access token."
  exit 1
fi

export GEN3_TOKEN_MAIN=$output
```

Approving anyway as it's not strictly necessary.

PR #179, opened 2019-10-28
test(Core Product): Executable test plan for gen3-client CLI utility

themarcelor:
PXP-4542

This executable test plan includes the following steps:

1. create dummy file (take sha)
2. gen3-client upload (store guid)
3. gen3-client single-download using guid (take sha)
4. compare sha's
5. delete file (_idempotent test_)

### DIFF
--- a/suites/portal/fileUploadTest.js
+++ b/suites/portal/fileUploadTest.js
@@ -1,10 +1,12 @@
 Feature('Data-Portal File Upload test with gen3-client CLI - https://ctds-planx.atlassian.net/browse/PXP-4542');
 
+// To be executed with GEN3_SKIP_PROJ_SETUP=true -- No need to set up program / retrieve access token, etc.
+
 const chai = require('chai');
 const {interactive, ifInteractive} = require('../../utils/interactive.js');
 const expect = chai.expect;
 const hostname = process.env.HOSTNAME;
-const profile = process.eng.NAMESPACE
+const profile = process.env.NAMESPACE
 
 Scenario('Create a folder in $vpc_name and generate a test file @manual', ifInteractive(
     async(I) => {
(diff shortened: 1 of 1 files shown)


suites/portal/fileUploadTest.js
```diff
@@ -0,0 +1,63 @@
...
+const chai = require('chai');
+const {interactive, ifInteractive} = require('../../utils/interactive.js');
+const expect = chai.expect;
+const hostname = process.env.HOSTNAME;
+const profile = process.env.NAMESPACE
+
+Scenario('Create a folder in $vpc_name and generate a test file @manual', ifInteractive(
+    async(I) => {
+        const result = await interactive (`
+            1. Login into $vpc_name
```
frickjack: Is $vpc_name the admin vm?  Why do we need to login to the admin vm to run this test?

suites/portal/fileUploadTest.js
```diff
@@ -0,0 +1,63 @@
...
+    expect(result.didPass, result.details).to.be.true;
+    }
+));
+
+// The test assumes the presence of a valid profile configuration in ~/.gen3/config (containing api-key, JWT token and api endpoint)
+Scenario('Run the gen3-client CLI utility and upload the test file to ${profile} @manual', ifInteractive(
+    async(I) => {
+        const result = await interactive (`
+            1. Run the gen3-cli upload command: gen3-client upload --profile=${profile} --upload-path=test.txt
+            2. Take note of the GUID that is printed as part of the output
```
frickjack: How about save the output of upload to a file?
```
gen3-cli upload ... | tee upload.log
```

Should step 3 be to `diff` the by-hand hash from above with the `gen3-cli upload` hash?

suites/portal/fileUploadTest.js
themarcelor: Good point. There's no need to do it.
This can be executed by any local terminal. Will add another commit shortly.

suites/portal/fileUploadTest.js
themarcelor: I've added the `diff` instruction as per your suggestion.

PR #180, opened 2019-10-28
PXP-4573 Test/arb profile

vpsx:
interactive test plan for configurable profile page

also add tests to arborist UI test plan for file page "download" button


vpsx:
feedback from Atharva: scenario title lines too long

PR #181, opened 2019-11-05
TestPlan : added for misconfig error and version checker

atharvar28:
This PR is adding test plan for gen3-client misconfiguration error check PXP-4571 and gen3-client version checker PXP-4570


suites/gen3-client/configCheckerTest.js
```diff
@@ -0,0 +1,56 @@
...
+        `);
+    expect(result.didPass, result.details).to,be.true;
+    }
+));
+
+Scenario('Wrong API key correct apiendpoint @manual', ifInteractive(
+    async(I) => {
+        const result = await interactive (`
+            1. user has a wrong cred.json (API key) and correct API endpoint
+            2. the misconfiguration checker should show 401 which means the API key for the user is not correct
```
mfshao: gen3-client won't display the HTTP error code in this case, instead it will display a message saying about `Invalid credentials for apiendpoint...`

suites/gen3-client/configCheckerTest.js
```diff
@@ -0,0 +1,56 @@
...
+        `);
+    expect (result.didPass, result.details).to,be.true;
+    }
+));
+
+Scenario('correct API key wrong apiendpoint @manual', ifInteractive(
+    async(I) => {
+        const result = await interactive (`
+            1. user has a correct cred.json (API key) but wrong API endpoint
+            2. the misconfiguration checker should show 404 or 405 which means user is hitting a wrong host
```
mfshao: similiar as above, the gen3-client won't display the HTTP error code in this case as well

atharvar28: @mfshao is the message displayed in this case same as 'Invalid credentials for apiendpoint...' or it displays that its the ' wrong apiendpoint '?

atharvar28: or else I can figure the exact message while testing! 😄

mfshao: in this case the message will be `The provided apiendpoint '${API_ENDPOINT}' is possibly not a valid Gen3 data commons` 

see https://github.com/uc-cdis/cdis-data-client/blob/master/gen3-client/g3cmd/configure.go#L45 and https://github.com/uc-cdis/cdis-data-client/blob/master/gen3-client/g3cmd/configure.go#L47 These are the only 2 cases the error message will be customized ;)

PR #184, opened 2019-11-21
chore(getAccessTokenInJS): Adding a js util script to fetch access tokens

themarcelor:
Tweaking the existing `getAccessToken.sh` script to call a new script: `getAccessToken.js`.
This new script invokes a new lib function: `getAccessTokenFromApiKey ` (which will also be used by executable tests).

Test run:
```
% /bin/bash getAccessToken.sh ~/.gen3/credentials.json
...
access_token: eyJ0eX***
```

### DIFF
--- a/getAccessToken.js
+++ b/getAccessToken.js
@@ -11,5 +11,5 @@ getAccessTokenFromApiKey(API_KEY, TARGET_ENVIRONMENT)
     .then((ACCESS_TOKEN) => {
 	console.log(ACCESS_TOKEN);
     }).catch((reason) => {
-	console.log('Failed: ' + reason.status +' - ' + reason.statusText);
+	console.log(`Failed: ${reason.status} - ${reason.statusText}`);
     });
--- a/getAccessToken.js
+++ b/getAccessToken.js
@@ -1,0 +1,14 @@
+const { getAccessTokenFromApiKey } = require('./utils/apiUtil');
+
+TARGET_ENVIRONMENT = process.env.GEN3_COMMONS_HOSTNAME;
+// console.log('target environment: ' + TARGET_ENVIRONMENT);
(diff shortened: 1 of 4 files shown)


getAccessToken.js
```diff
@@ -0,0 +1,14 @@
...
+
+TARGET_ENVIRONMENT = process.env.GEN3_COMMONS_HOSTNAME;
+// console.log('target environment: ' + TARGET_ENVIRONMENT);
+API_KEY = process.env.API_KEY;
+// console.log('api key: ' + API_KEY);
+
+// obtain access token through fence API
+getAccessTokenFromApiKey(API_KEY, TARGET_ENVIRONMENT)
+    .then((ACCESS_TOKEN) => {
+	console.log('access_token: ' + ACCESS_TOKEN);
```
frickjack: Just output the token, so we can do things like: `token="$(getAccessToken.sh)"`

getAccessToken.js
```diff
@@ -0,0 +1,14 @@
+const { getAccessTokenFromApiKey } = require('./utils/apiUtil');
+
+TARGET_ENVIRONMENT = process.env.GEN3_COMMONS_HOSTNAME;
+// console.log('target environment: ' + TARGET_ENVIRONMENT);
+API_KEY = process.env.API_KEY;
+// console.log('api key: ' + API_KEY);
+
```
frickjack: Maybe add a comment to point out that command line argument processing is handled in `getAccessToken.sh` ...

getAccessToken.js
themarcelor: That's clever.
Pushing a commit to address that.

getAccessToken.js
themarcelor: Good call. Pushing the change now.

PR #195, opened 2019-12-03
chore(integration): Failing fast when google svc account cant be deleted

themarcelor:
Looks like we are not failing-fast when it comes it some Google-service-account tests.

Here's one of the util classes we use in the integration tests:
```
  async deleteServiceAccount(serviceAccountID, projectID = null) {
...
      cloudResourceManager.projects.serviceAccounts.delete(request, (err, res) => {
        if (err) {
          resolve(Error(err));
          return;
        }
        resolve(res.data);
      });
    }));
  },
```
which is called from `suites/google/googleServiceAccountKeyTest.js`:
```
  await google.deleteServiceAccountKey(key.name);
```
even if the function returns an Error object, we don't handle that.

Which leads to failures in the subsequent steps.

We need to fail fast to speed up the debugging process in case of any intermittent issues.

### DIFF
--- a/suites/google/googleServiceAccountKeyTest.js
+++ b/suites/google/googleServiceAccountKeyTest.js
@@ -231,9 +231,10 @@ Scenario('SA key removal job test: remove expired creds that do not exist in goo
   let credsList = getCredsRes.access_keys;
   let key = credsList.filter(key => key.name.includes(credsKey1))[0];
   const deletionResult = await google.deleteServiceAccountKey(key.name);
-    chai.expect(
-	deletionResult instanceof Error, `Google service account deletion returned an Error: ${deletionResult}`
-    ).to.equal(false);
+  chai.expect(
+    deletionResult instanceof Error,
+    `Error during Google service account deletion: ${deletionResult}`
+  ).to.equal(false);
 
   // Wait for the keys to expire
   console.log('waiting for the key to expire');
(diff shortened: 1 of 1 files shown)


suites/google/googleServiceAccountKeyTest.js
```diff
@@ -230,7 +230,10 @@ Scenario('SA key removal job test: remove expired creds that do not exist in goo
   let getCredsRes = await fence.do.getUserGoogleCreds(users.user0.accessTokenHeader);
   let credsList = getCredsRes.access_keys;
   let key = credsList.filter(key => key.name.includes(credsKey1))[0];
-  await google.deleteServiceAccountKey(key.name);
+  const deletionResult = await google.deleteServiceAccountKey(key.name);
+    chai.expect(
+	deletionResult instanceof Error, 'Google service account deletion must not return an Error'
```
paulineribeyre: debugging would be easier if you included the error's description in the error message returned by `expect`... if it's possible to grab it from `deleteServiceAccountKey()`

themarcelor: Brilliant suggestion @paulineribeyre ! I will push another change.

paulineribeyre: i'm not sure at returns the errors as it is https://github.com/uc-cdis/gen3-qa/blob/aa222a5c2c8c26a82b886a601c3e6b40905695e3/utils/google.js#L398-L400

themarcelor: NICE TEAM WORK! 
#oneTeam!

PR #199, opened 2019-12-06
 task(dcf): Executable tests for DCF: IndexD scenarios

themarcelor:
Sixth set of executable tests for DCF.

List of DCF Staging test scenarios (as per https://github.com/uc-cdis/cdis-wiki/blob/master/qa/reports/2019/11/DcfStaging-20191114.md):

1. Bootstrapping accounts (linking Google account)
2. User Managed Service Accounts
3. Google Credentials
4. Signed URLs
5. OAuth2/OIDC-flow/Client-Testing
6. **--> Indexd**

### DIFF
--- a/suites/pre-release/dcf/6.indexdTest.js
+++ b/suites/pre-release/dcf/6.indexdTest.js
@@ -18,19 +18,17 @@ BeforeSuite(async(I) => {
     console.log('Setting up dependencies...');
     // making this data accessible in all scenarios through the actor's memory (the "I" object)
     I.cache = {};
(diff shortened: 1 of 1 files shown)


haraprasadj (approved):

haraprasadj:
We may not need to add await when using then(). Stackoverflow says there is no harm in doing it though, so it is your call :) (https://stackoverflow.com/questions/55019621/using-async-await-and-then-together)

haraprasadj (approved):

PR #206, opened 2019-12-13
chore(logging): Move logging line to keep it within its test scenario

themarcelor:
CodeceptJS runs this line even before the suites/apis/dataStageOIDCFlowTest.js Feature+Scenarios are executed, which leads to some misleading analysis while debugging some test failure.


themarcelor:
Will sort this out in one of the other PRs

PR #215, opened 2020-01-07
Doc(reporting): QA Reports landing page + instructions

themarcelor:
This PR contains the static assets to render the QA reports landing page.
Now our internal users will be able to easily find all the reports.


themarcelor:
We're moving this logic to the `cloud-automation` repo.

PR #217, opened 2020-01-10
chore:adding smoke test suite

atharvar28:
This PR is to add a smoke test suite into the gen3-qa repo. Also includes executable smoke test plan for brain commons promotion

### DIFF
--- a/suites/smokeTests/brainTests.js
+++ b/suites/smokeTests/brainTests.js
@@ -85,7 +85,7 @@ Scenario('Version Checker error @manual @core', ifInteractive(
   async () => {
     const result = await interactive(`
             1. After the successful installation and configuration of profile, user can use gen3-client command on terminal console
-            2. the version checker will show 'A new version of gen3-client is avaliable! The latest version is ${LATEST_VERSION}. You are using version ${CURRENT_VERSION}'
+            2. the version checker will show 'A new version of gen3-client is avaliable! The latest version is LATEST_VERSION. You are using version CURRENT_VERSION'
             Please download the latest gen3-client release from https://github.com/uc-cdis/cdis-data-client/releases/latest' message on the console if a newer version of gen3-client is available
 
             Note : This test can be done only locally currently as there are no versions for gen3-client. To carry out the test locally, follow this https://github.com/uc-cdis/cdis-data-client#installation and make a version change in 'gitversion' on path gen3-client/g3cmd/gitversion.go
--- a/suites/smokeTests/brainTests.js
+++ b/suites/smokeTests/brainTests.js
@@ -113,7 +113,7 @@ Scenario('Configuring gen3-client @manual @core', ifInteractive(
 
 // misconfiguration error checker
 Scenario('Wrong API key correct apiendpoint @manual @core', ifInteractive(
-  async (I) => {
+  async () => {
     const result = await interactive(`
             1. user has a wrong cred.json (API key) and correct API endpoint
             2. the misconfiguration checker displays a message Invalid credentials for apiendpoint '<apiendpoint>': check if your credentials are expired or incorrect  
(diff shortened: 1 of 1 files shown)


suites/smokeTests/brainTests.js
```diff
@@ -0,0 +1,200 @@
+Feature("Brain Promotion Smoke Test Suite");
+
+const chai = require('chai');
+const {interactive, ifInteractive} = require('../../utils/interactive.js');
+const expect = chai.expect;
+
+//@toBeAutomated
+Scenario('Login with Synapse first time @manual', ifInteractive(
```
themarcelor: @atharvar28 , just a quick heads-up:
Since you are picking the specific set of executable tests that comprise Brain's smoke tests, I think we should leverage the 'at' (@) notation (CodeceptJS tag) to facilitate the execution of the test suite.

I have to do the same for DataSTAGE and DCF, a per:
> PXP-3839: datastage smoke test suite
> PXP-4465: dcf smoke test suite

Check the ticket's description.

themarcelor: So maybe we could use `@brain` here? So it would contain both `@manual @brain` ?
What do you think?

atharvar28: I am more inclined towards using `@brainsmoke` to make a clear indication of what is happening in a particular test run @themarcelor - just a thought! The annotations to test scenarios can keep going up and it would be difficult to come back once the process of annotations is on full throttle. but we can discuss later!

themarcelor: awesome. And for common / overlapping ones we can annotate as @core 
@haraprasadj , what do you reckon?

haraprasadj: we can just have @smoke, @brain, @dcp etc. Some tests will have multiple annotations (@brain and @smoke) then we can test all kinds of combinations - all brain tests, smoke tests for brain etc.

themarcelor: That also makes sense...
so it would be like:
```
`@brain @smoke`
```
And then we just grep whatever we want:
e.g.,
```
% RUNNING_LOCAL=true NAMESPACE="jenkins-genomel" npm test -- --verbose --grep '@smoke|@dcf'

> gen3-qa@0.1.0 test /Users/marcelocostarodrigues/workspace/gen3-qa
> codeceptjs run --plugins allure "--verbose" "--grep" "@smoke|@dcf"

jenkins-genomel
true
Running against k8s cluster.
Start setup
NAMESPACE: jenkins-genomel
HOSTNAME: jenkins-genomel.planx-pla.net
Tags:
[ '@smoke', '@dcf' ]
```

see?

[ '@smoke', '@dcf' ]
--

suites/smokeTests/brainTests.js
```diff
@@ -0,0 +1,198 @@
...
+            2. the user sees 'Create API key' button and table with resource permissions
+            3. user clicks on 'Create API key' button and pop-up with API key created for the user
+            4. user can download the 'credential.json' to carry out various other test execution against https://gen3qa.braincommons.org/
+            5. the user is able see the project access table with right permissions to the right projects
+        `);
+    expect(result.didPass, result.details).to.be.true;
+  },
+));
+
+Scenario('Install gen3-client @manual @core', ifInteractive(
```
themarcelor: Didn't we agree to replace @core with @smoke ?
All scenarios should contain 3 annotations:
```
@manual @smoke @brain
```

Right?

suites/smokeTests/brainTests.js
```diff
@@ -0,0 +1,198 @@
...
+            3. the user is redirected to Synapse site, click on 'Allow' which redirects to https://gen3qa.braincommons.org/
+            4. After being re-directed, there is a 'Terms of Use and Privacy Policy' pop-up on https://gen3qa.braincommons.org/ 
+            5. the user have to hit 'Agree' two times to get through to the data commons
+            6. Log out of Synapse account
+        `);
+    expect(result.didPass, result.details).to.be.true;
+  },
+));
+
+Scenario('Login with Synapse @manual @dream', ifInteractive(
```
themarcelor: Looks like you missed this one?

PR #218, opened 2020-01-15
Fix: python 3 for data-simulator

paulineribeyre:
Fixed an encoding issue as part of https://github.com/uc-cdis/data-simulator/pull/35 that makes the data-simulator non py2-compatible

### Bug Fixes
- The data-simulator is now python 3, the test pipeline should run it with python 3


paulineribeyre:
ignoring failed jenkins tests and merging - the failures are urelated and i reran many times

PR #222, opened 2020-01-16
chore(qa): Deleting all SA keys before test scenarios + Refactoring and additional debugging to investigate intermittent issues

themarcelor:
Refactoring code, introducing additional debugging lines and retries.

Also deleting SA keys before each scenario.

### DIFF
--- a/services/apis/fence/fenceSequences.js
+++ b/services/apis/fence/fenceSequences.js
@@ -126,7 +126,7 @@ module.exports = {
       const response = await fenceTasks.createTempGoogleCreds(accessTokenHeaders, expires_in);
       // response from creating temporary Google credentials does not have
       // nested property data.private_key (which means we didn't get back valid Google credentials
-      if (!myBadJson.hasOwnProperty('data') || !myBadJson.data['private_key']) {
+      if (!response.hasOwnProperty('data') || !response.data['private_key']) {
         console.log(`Failed to create google temp creds on attempt ${i}:`);
         console.log(`Invalid response: ${JSON.stringify(response)}`);
         await sleepMS(1 * 1000);
--- a/services/apis/fence/fenceSequences.js
+++ b/services/apis/fence/fenceSequences.js
@@ -128,7 +128,7 @@ module.exports = {
       // nested property data.private_key (which means we didn't get back valid Google credentials
       if (!response.hasOwnProperty('data') || !response.data['private_key']) {
         console.log(`Failed to create google temp creds on attempt ${i}:`);
-        console.log(`Invalid response: ${response.data}`);
+        console.log(`Invalid response: ${JSON.stringify(response.data)}`);
         await sleepMS(2 * 1000);
         if (i === nAttempts - 1) {
           throw new Error('Failed create temp google creds!');
--- a/services/apis/fence/fenceSequences.js
+++ b/services/apis/fence/fenceSequences.js
@@ -121,11 +121,22 @@ module.exports = {
    * @returns {Promise<Gen3Response>}
    */
   async createTempGoogleCreds(accessTokenHeaders, expires_in = null) {
-    const response = await fenceTasks.createTempGoogleCreds(accessTokenHeaders, expires_in);
-    expect(response,
-      'response from creating temporary Google credentials does not have nested '
-      + 'property data.private_key (which means we didn\'t get back valid Google credentials)').has.nested.property('data.private_key');
-    return response;
+    const nAttempts = 3;
+    for (let i = 0; i < nAttempts; i += 1) {
+      const response = await fenceTasks.createTempGoogleCreds(accessTokenHeaders, expires_in);
+      // response from creating temporary Google credentials does not have
+      // nested property data.private_key (which means we didn't get back valid Google credentials
+      if (!myBadJson.hasOwnProperty('data') || !myBadJson.data['private_key']) {
+        console.log(`Failed to create google temp creds on attempt ${i}:`);
+        console.log(`Invalid response: ${JSON.stringify(response)}`);
+        await sleepMS(1 * 1000);
+        if (i === nAttempts - 1) {
+          throw new Error('Failed create temp google creds!');
+        }
+      } else {
+        return response;
+      }
+    }
   },
 
   /**
--- a/services/apis/fence/fenceSequences.js
+++ b/services/apis/fence/fenceSequences.js
@@ -131,7 +131,7 @@ module.exports = {
         console.log(`Invalid response: ${JSON.stringify(response.data)}`);
         await sleepMS(2 * 1000);
         if (i === nAttempts - 1) {
-          throw new Error('Failed to create temp google creds due to: ${JSON.stringify(response.data)}! Num of attempts: ${i}.');
+          throw new Error(`Failed to create temp google creds due to: ${JSON.stringify(response.data)}! Num of attempts: ${i}.`);
         }
       } else {
         return response;
--- a/services/apis/fence/fenceSequences.js
+++ b/services/apis/fence/fenceSequences.js
@@ -1,3 +1,4 @@
+/*eslint-disable */
 const chai = require('chai');
 
 const { expect } = chai;
@@ -36,6 +37,7 @@ module.exports = {
    */
   async linkGoogleAcctMocked(userAcct, expires_in = null) {
     const linkRes = await fenceTasks.linkGoogleAcctMocked(userAcct, expires_in);
+    console.log(`### ## linkRes for [${userAcct.username}]: ${linkRes}`);
     fenceQuestions.mockedLinkSuccess(linkRes);
     return linkRes;
   },
--- a/services/apis/fence/fenceSequences.js
+++ b/services/apis/fence/fenceSequences.js
@@ -128,8 +128,8 @@ module.exports = {
       // nested property data.private_key (which means we didn't get back valid Google credentials
       if (!response.hasOwnProperty('data') || !response.data['private_key']) {
         console.log(`Failed to create google temp creds on attempt ${i}:`);
-        console.log(`Invalid response: ${JSON.stringify(response)}`);
-        await sleepMS(1 * 1000);
+        console.log(`Invalid response: ${response.data}`);
+        await sleepMS(2 * 1000);
         if (i === nAttempts - 1) {
           throw new Error('Failed create temp google creds!');
         }
--- a/services/apis/fence/fenceSequences.js
+++ b/services/apis/fence/fenceSequences.js
@@ -8,7 +8,7 @@ chai.config.truncateThreshold = 0;
 const fenceQuestions = require('./fenceQuestions.js');
 const fenceTasks = require('./fenceTasks.js');
 const fenceProps = require('./fenceProps.js');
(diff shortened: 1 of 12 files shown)


paulineribeyre:
https://github.com/uc-cdis/gen3-qa/blob/master/suites/google/googleDataAccessTest.js#L33-L40

services/apis/fence/fenceSequences.js
```diff
@@ -35,9 +36,22 @@ module.exports = {
    * @returns {Promise<Gen3Response>}
    */
   async linkGoogleAcctMocked(userAcct, expires_in = null) {
-    const linkRes = await fenceTasks.linkGoogleAcctMocked(userAcct, expires_in);
-    fenceQuestions.mockedLinkSuccess(linkRes);
-    return linkRes;
+    const nAttempts = 3;
+    for (let i = 0; i < nAttempts; i += 1) {
```
Avantol13: the testing framework has a generic `.retry()` you can append to a given test, this is fine for more granular retries, but we can just retry the entire test too if we're not already

services/apis/fence/fenceTasks.js
```diff
@@ -217,17 +219,19 @@ module.exports = {
     }
 
     let res = await getNoRedirect(url, headers);
+    // console.log(`### NoRedirectURL Res: ${res.data}`);
```
Avantol13: if this is useful in the future can you uncomment? otherwise I'd prefer to just completely remove the line (same for other debugging lines that are commented out)

suites/apis/OAuth2Test.js
```diff
@@ -24,21 +25,21 @@ Scenario('Authorization code flow: Test that successfully generates code', async
     fence.props.clients.client.id, 'code', 'openid+user',
   );
   fence.ask.assertContainSubStr(resULR, ['code=']);
-});
+}).retry(2);
```
Avantol13: oh :man_facepalming: this is the retry I was talking about in a previous comment. :+1:

suites/apis/exportToWorkspaceTest.js
```diff
@@ -1,8 +1,7 @@
-Feature('ExportToWorkspaceAPITest');
+Feature('ExportToWorkspaceAPITest').retry(2);
```
Avantol13: are you sure this works? You can retry on a Feature instead of individual Scenarios?

suites/google/googleDataAccessTest.js
```diff
@@ -75,31 +77,269 @@ BeforeSuite(async (indexd) => {
...
+      };
+    }
+  });
+});
+
 After(async (fence, users) => {
   // Cleanup after each scenario
-  const unlinkResults = Object.values(users).map((user) => fence.do.unlinkGoogleAcct(user));
+  const unlinkResults = Object.values(users).map(async(user) => {
+    await apiUtil.sleepMS(1 * 1000).then(() => {
```
Avantol13: why the sleep? I _should_ be okay without it I think

suites/google/googleDataAccessTest.js
```diff
@@ -75,31 +77,269 @@ BeforeSuite(async (indexd) => {
...
+    let User0signedUrlQA1Res = '';
+    let tempCreds0Res = '';
+
+    // Cannot retry more tha 10 times due to SA key limit (10)
+    // FYI: All Keys are cleaned up in the Before() block
+    const nAttempts = 8;
+    for (let i = 0; i < nAttempts; i += 1) {
+      console.log('make sure google account user0 is unlinked');
+      const unlinkResult = await fence.complete.forceUnlinkGoogleAcct(users.user0);
+      await apiUtil.sleepMS(1 * 1000);
```
Avantol13: the above should only return once it's complete, I'm not sure the sleep is absolutely necessary in this case

suites/google/googleDataAccessTest.js
```diff
@@ -75,31 +77,269 @@ BeforeSuite(async (indexd) => {
...
+      console.log(`Use User0 to create signed URL for file in QA. Attempt #${i}`);
+      User0signedUrlQA1Res = await fence.do.createSignedUrlForUser(
+          indexed_files.qaFile.did, users.user0.accessTokenHeader,
+      );
+
+      console.log(`User0signedUrlQA1Res: ${JSON.stringify(User0signedUrlQA1Res)}`);
+      User0signedUrlQA1FileContents = await fence.do.getFileFromSignedUrlRes(
+        User0signedUrlQA1Res,
+      );
+      console.log(`The friggin\' contents of the QA file: ${stringify(User0signedUrlQA1FileContents).substring(User0signedUrlQA1FileContents.length-100, User0signedUrlQA1FileContents.length)}`);
```
Avantol13:
```suggestion
      console.log(`The contents of the QA file: ${stringify(User0signedUrlQA1FileContents).substring(User0signedUrlQA1FileContents.length-100, User0signedUrlQA1FileContents.length)}`);
```

suites/google/googleServiceAccountKeyTest.js
```diff
@@ -18,27 +19,27 @@ function calculateSAKeyAge(creationDate) {
 }
 
 BeforeSuite(async (google, fence, users) => {
-  console.log('cleaning up old keys from the service accounts in the dcf-integration GCP project');
-  const dcfSvcAccounts = await google.listServiceAccounts('dcf-integration');
-  // console.log('#### ##: ' + dcfSvcAccounts);
-  dcfSvcAccounts.forEach(async (svcAccount) => {
-    const saName = svcAccount.email;
+  console.log('cleaning up old keys from the user0 service account in the dcf-integration GCP project');
+  const getCredsRes = await fence.do.getUserGoogleCreds(users.user0.accessTokenHeader);
```
Avantol13: this is only cleaning things up for user0, before it was cleaning up for all SAs. where do you clean up user1/user2 and other SAs (or you rely on other test suites/features to do that)?

suites/google/googleServiceAccountKeyTest.js
```diff
@@ -235,6 +247,12 @@ Scenario('SA key removal job test: remove expired creds @reqGoogle', async (fenc
   console.log('Clean up expired Service Account keys');
   bash.runJob('google-manage-keys-job');
```
Avantol13: is this another instance where we need to remove the `-job`?

suites/google/googleServiceAccountKeyTest.js
```diff
@@ -251,7 +269,14 @@ Scenario('SA key removal job test: remove expired creds @reqGoogle', async (fenc
   await fence.do.deleteTempGoogleCreds(
     creds0Key,
     users.user0.accessTokenHeader,
-  );
+  ).then((deletionResult) => {
+    console.log(`deletionResult: ${JSON.stringify(deletionResult)}`);
+    console.log(`is it an error?: ${deletionResult instanceof Error}`);
+    chai.expect(
```
Avantol13: if this fails here, the `files.deleteFile(` below won't run. this is probably fine but just a note that the first failing assert ends the scenario

services/apis/fence/fenceSequences.js
themarcelor: I see you've commented in another line that contains the `.retry()` function.
That's right, this is just a granular one because, during the initial experiments with this PR, this very first test was failing most often so another objective I have with these changes is to *fail fast*, hence, the retries on this first check.

services/apis/fence/fenceTasks.js
themarcelor: This is a left-over from the initial debugging.
We can actually access this information through the `--verbose` flag when we run `npm test`.

I'm happy to remove the line, however, I'd prefer to do it in a new PR 😄 (I want to merge this and monitor the results of other people's PRs -- I want to see how these changes behave out in the world and tweak / revert accordingly).

suites/apis/OAuth2Test.js
themarcelor: Correct.
Once we apply the mitigation as the first call to action, we should try to increase the observability around such retries and deliberate further on whether or not we want to keep them or:
- Perform further low level investigation on why the system produces invalid results intermittently.
- Make the tests smarter by avoiding garbage from previous runs that haven't been cleaned-up properly,  some smart sleep calls to eliminate potential race conditions, etc.

suites/apis/exportToWorkspaceTest.js
themarcelor: It does work. Here's the official doc:
https://codecept.io/basics/#retry-feature

suites/google/googleDataAccessTest.js
themarcelor: Yeah, that was mostly done based on dogmatic conjectures around what would happen in the background if Fence, Database, Google, et al. take too long to process this sort of command.

Again, I'm happy to remove it in a new PR but I'd like to see how this one behaves out in the wild.

suites/google/googleDataAccessTest.js
themarcelor: Same situation from this other comment:
https://github.com/uc-cdis/gen3-qa/pull/222/files/f21b4713a85bc1cfd2c2cce16b8dad5da75e6f90#r369252285

suites/google/googleDataAccessTest.js
themarcelor: Oh, a bit of colloquialism there 😄 
Is it ok to remove it in a new PR? I'm worried about hitting some other obscure intermittent issues and start the yak shaving all over again (> 600 lines).

suites/google/googleServiceAccountKeyTest.js
themarcelor: `googleServiceAccountKeyTest.js` only cares about keys from `user0`. So these are the only ones I'm cleaning up.

`googleDataAccessTest.js` is the one that cares about all 3.

BTW, I'm aware there's some obvious code repetition there that must be refactored / moved into an `utils / commons` function.  

There are many conjectures involving the keys' clean-up process.
I just want to start with a clean slate right before each test starts.
Maybe we could do it only once in some other script (e.g., `test_setup.js`) or maybe clean up everything in the Groovy pipeline "Post" stage? Regardless of the ideal solution, this mitigation might help us get unblocked for now so we can decide/design/implement a proper solution later.

suites/google/googleServiceAccountKeyTest.js
themarcelor: I have this tiny change in a separate PR that miraculously passed:
https://github.com/uc-cdis/gen3-qa/pull/219/files

suites/google/googleServiceAccountKeyTest.js
themarcelor: We clean-up workspaces on every test run so the Jenkins file system for that PR run is wiped out.
That is, however, a real issue for whoever is running this locally (But, in case that happens, they would see some weird file in their `git status` output so... there is a manual workaround for this edge case).

Avantol13 (approved):
Approved with caveats:

* would like to see another PR for the small fixes we avoided addressing here b/c we don't want to wait for more flaky tests to pass

And a final note for posterity:

* this *will* make the integration tests take longer b/c of the new "sleeps" and splitting of the google data access test (b/c usersync will run multiple times now)

PR #246, opened 2020-02-13
changed login test to support top bar Login button

haraprasadj:
### Improvements
Updated login test to use the Login button added to top bar

### DIFF
--- a/services/portal/home/homeTasks.js
+++ b/services/portal/home/homeTasks.js
@@ -20,6 +20,7 @@ module.exports = {
   login(username) {
     this.goToHomepage();
     I.setCookie({ name: 'dev_login', value: username });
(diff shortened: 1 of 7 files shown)


services/portal/home/homeTasks.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
   login(username) {
     this.goToHomepage();
     I.setCookie({ name: 'dev_login', value: username });
+    portal.clickProp(homeProps.loginButton);
```
mfshao: this will fail if a commons is using an older portal?

haraprasadj: added a flag to handle portal versions on which to run the test

PR #251, opened 2020-03-04
added test cases for sheepdog admin

haraprasadj:


test_list_auto.txt
```diff
@@ -0,0 +1,186 @@
+Tests from /Users/haraprasadjuvvala/workspace/gen3/gen3-qa:
```
themarcelor: This might've been added to the PR by accident.

haraprasadj: It was intentional. Let it get added just for informational purposes. We can keep updating periodically. Also, did not want to raise a separate PR just for this :)

themarcelor: Best to keep separate PRs for different purposes to help with traceability (so we can glance at the title and know what the change is about).

But if you really want to deliver it through this PR.. meh. Not a big deal.

haraprasadj: True, but this is just a "good-to-have" info. Also, the doc-only tag did not support txt files when I created this PR, so separate PR meant running all integration tests and debugging their failures if any. Overkill for a file that just has a listing of test cases.

PR #266, opened 2020-03-17
task(load test): Introduce load test for Metadata Service

themarcelor:
Jira Ticket: [PXP-5603](https://ctds-planx.atlassian.net/browse/PXP-5603)

This PR contains the load test script for the new Metadata Service + adjustments to the load testing framework to generate a `uuid.js` module on the fly and drop the file under `load-testing/libs`.

The load test generates 2 GUIDs for each Virtual User's iteration and performs CREATE, QUERY and DELETE operations.


load-testing/load-test-basic-auth-sample.json
```diff
@@ -0,0 +1,139 @@
...
+      "a": 1
+    },
+    "fictitiousRecord2": {
+      "nestedData": {
+        "b": 2
+      }
+    }
+  },
+  "basic_auth": {
+    "username": "<insert_username_here>",
```
haraprasadj: should we change these to env variables?

themarcelor: The JSON is a sample and it is meant to be manually tailored by the user.
There is no mechanism to seamlessly inject an environment variable into the JSON.

We could.however, in future versions, let the user decide whether the secret is hardcoded in there or have something like ENV:MY_VARIABLE and then we can parse the JSON and try to fetch the corresponding variable whenever we find the `ENV:` string. Something like that.

haraprasadj: sounds good

PR #271, opened 2020-03-20
added elasticsearch query util

haraprasadj:


suites/elasticTest.js
```diff
@@ -0,0 +1,21 @@
...
+
+const { expect } = require('chai');
+const { search } = require('../utils/elasticsearch.js');
+
+const index = 'etl_mickey_1';
+const query = '{ "query": { "nested": { "path": "BCMAMedicationLog", "query": { "range": { "BCMAMedicationLog.DosesOrdered": { "lt": 0.0029 } } } } } }';
+
+// Sample scenarios to query elasticsearch and validate results
+// scenario marked as manual to prevent it from running in CI pipeline
+Scenario('check length', async () => {
```
themarcelor: Should we favour this new `.tag` chained function over the old way of tagging?

utils/elasticsearch.js
```diff
@@ -0,0 +1,19 @@
+const { Client } = require('@elastic/elasticsearch');
+const { Bash } = require('./bash');
+
+const service = 'aws-es-proxy-deployment';
+const bash = new Bash();
+const ES_URL = 'http://localhost:9200';
+
+module.exports = {
+  search: async (index, query) => {
+    const res = bash.runCommand(`g3kubectl port-forward deployment/${service} 9200:9200 &`);
```
themarcelor: Maybe get a +1 from DevOps to gather caveats on potential accumulation of kubectl port-forward processes in the background? 🤷‍♂

suites/elasticTest.js
haraprasadj: I thought it looked nice :)

suites/elasticTest.js
```diff
@@ -0,0 +1,21 @@
+Feature('es');
+
+const { expect } = require('chai');
+const { search } = require('../utils/elasticsearch.js');
+
+const index = 'etl_mickey_1';
```
themarcelor: Was that supposed to be hardcoded?

utils/elasticsearch.js
haraprasadj: it should only run once - you should get an error if you retry - Error listen tcp6 [::1]:9200: bind: address already in use]

suites/elasticTest.js
haraprasadj: this sample test is temporary anyway

utils/elasticsearch.js
themarcelor: Concurrent PR checks + potential executions from local terminals can create multiple instances of the same command, right?

suites/elasticTest.js
haraprasadj: Again, sample test ... once we start using it, we can remove this

utils/elasticsearch.js
haraprasadj: The command will fail to run a second time.

PR #273, opened 2020-03-20
chore(qa): Apply workaround to avoid running DRS tests for March Rele…

themarcelor:
temporary patch


themarcelor:
Closing this in favour of a better solution.

PR #308, opened 2020-04-29
Task/tests for indexing gui and sower jobs

themarcelor:
Jira Ticket: [PXP-5786](https://ctds-planx.atlassian.net/browse/PXP-5786)

This PR comprises 2 new automated test scenarios for the Indexing GUI `/indexing` & the recently-added sower jobs that allow the users to:
1. Upload and register a set of indexd records through a manifest and 
2. Generate and download a manifest containing a full list of indexd records from the Gen3 Commons environment.

This was tested against qa-dcp and all the required changes have also been applied to Jenkins CI environments. It has also been executed manually against `jenkins-genomel`, which presented positive results.

### New Features
New automated test with the tag @indexing.

### Improvements
Adding condition to `run-tests.sh` to only execute the tests if the `sower-manifest` configmap contains the relevant sower jobs declared in there. This is a temporary measure until the invalid URLs can redirect  the user to a proper 404 Not Found page.

### tests

********
TEST: Navigate to the indexing page and upload a test manifest @indexing
RESULT: passed
RETRIES: 0
TIMESTAMP: Wed Apr 29 2020 16:25:55 GMT-0500 (Central Daylight Time)
********
  ✔ OK in 86868ms

********
TEST: Navigate to the indexing page and download a full indexd manifest @indexing
RESULT: passed
RETRIES: 0
TIMESTAMP: Wed Apr 29 2020 16:26:45 GMT-0500 (Central Daylight Time)
********
  ✔ OK in 49398ms

### DIFF
--- a/suites/portal/indexingPageTest.js
+++ b/suites/portal/indexingPageTest.js
@@ -1,11 +1,12 @@
 /*
+ Indexing GUI & Indexing/Manifest Sower Jobs (PXP-5786)
  This test plan has a few pre-requisites:
  1. Sower must be deployed and
  2. The environment's manifest must have the indexing jobs declared
     within the sower config block (manifest-indexing & indexd-manifest)
  3. The Indexing GUI is only available in data-portal >= 2.24.9
 */
-Feature('Indexing page - PXP-5786');
+Feature('Indexing GUI');
 
 const { expect } = require('chai');
 const { sleepMS, Gen3Response } = require('../../utils/apiUtil.js');
--- a/suites/portal/indexingPageTest.js
+++ b/suites/portal/indexingPageTest.js
@@ -8,6 +8,8 @@
 Feature('Indexing page - PXP-5786');
 
 // const stringify = require('json-stringify-safe');
+const path = require("path")
+const fs = require("fs")
 const { expect } = require('chai');
 const { sleepMS, Gen3Response } = require('../../utils/apiUtil.js');
 const { Bash } = require('../../utils/bash.js');
@@ -89,6 +91,7 @@ Scenario('Navigate to the indexing page and upload a test manifest @indexing', a
 
     if (indexdRecordRes.data.hashes) {
       expect(indexdRecordRes.data.hashes.md5).to.equal(testHash);
+      break
     } else {
       console.log(`WARN: The indexd record has not been created yet... - attempt ${i}`);
       await sleepMS(5000);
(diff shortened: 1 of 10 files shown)


suites/portal/indexingPageTest.js
```diff
@@ -0,0 +1,136 @@
...
+    manifestDownloadUrl.toString(),
+  ).then((res) => new Gen3Response(res));
+  console.log(`### downloadOuput: ${JSON.stringify(getManifestRes)}`);
+
+  const testManifestData = contentsOfTestManifest.split('\n')[1];
+  const downloadedManifestData = getManifestRes.body;
+
+  expect(
+    downloadedManifestData,
+  ).to.include(
```
ZakirG, line 133: if the BeforeSuite is deleting all indexd records, this test should still pass if you use the stricter "equal" instead of "include" yeah?

test_setup.js
```diff
@@ -250,6 +250,12 @@ module.exports = async function (done) {
     assertEnvVars(basicVars.concat(googleVars, submitDataVars));
     console.log('TEST_DATA_PATH: ', process.env.TEST_DATA_PATH);
 
+    // If testedEnv is not defined, infer FQDN based on the NAMESPACE
+    if (process.env['testedEnv'] === '' || process.env['testedEnv'] === undefined) {
```
ZakirG, line 254: i dont think javascript style encourages use of the triple equals against the undefined keyword. i think the string comparison like so is preferred: 
`typeof process.env['testedEnv'] === 'undefined'`

ZakirG: source: https://stackoverflow.com/questions/3390396/how-to-check-for-undefined-in-javascript

suites/portal/indexingPageTest.js
themarcelor: The 1st Scenario creates the indexd record and the 2nd Scenario downloads the manifest and makes sure the same data is found (GUID, md5 hash, size, url, etc.).
CodeceptJS `BeforeSuite` runs before all scenarios so that doesn't impact the 2nd test scenario.
There's a separate `Before` function that runs before every Scenario (which behaves similarly to Python's unittest `SetUp()` function), that one would totally mess this up, but that is not used in this script.

test_setup.js
themarcelor: Oh, I was reproducing _prior art_ found on the same script -> Line 17:
```
const inJenkins = (process.env.JENKINS_HOME !== '' && process.env.JENKINS_HOME !== undefined);
```

But I can push another commit for that, np.

suites/portal/indexingPageTest.js
ZakirG: would the test pass if you used "equal" instead of "include" ? i feel like a stricter condition is preferable ~

ZakirG: are there other records or might there be duplicates?

themarcelor: I'm using include because the test `manifest.tsv` file also becomes an indexd record (but I only care about the actual record that is created through the sower job, not the one that is created by `ssjdispatcher`).
The `BeforeSuite` should delete 'em all so we don't expect other records.
This test will definitely pass if there are duplicates, that is a good point.
Is it realistic to expect the sower job to create duplicates though??

PR #350, opened 2020-06-04
added test cases for 404 page

haraprasadj:


haraprasadj:
> should this be an interactive test? i think this test can be automated
Yeah, but I don't know the properties of the 404 page, so will make it automated once I know


> also i think it would be worth to test the following scenarios:
> 
> 1. if i'm trying to navigate to some predefined location, like `/profile`, i should NOT be directed to the 404 page.
> 2. if i'm trying to navigate to a project page that EXISTS in the env, like `/DEV-test`, i should NOT be directed to the 404 page, but i should be redirected to the project submission page for this specific project.

I will add these scenarios, thanks!

PR #362, opened 2020-06-12
added tests for data in explorer page

haraprasadj:


suites/portal/dataExplorerTest.js
```diff
@@ -101,8 +101,82 @@ Scenario('Faceted search - Nested data - Nested fields can be shown @manual', if
...
+          }
+          Oncology_Primary {
+            Multiplicitycounter
+            ICDOSite
+          }
+        }
+      }
+      3. Navigate to the explorer page (e.g. https://qa-mickey.planx-pla.net/explorer)
+      4. Verify that the data in the table matches the data from the query page
+      5. Verify that there is a sub-table for each nested parent (ActionableMutations and Oncology_Primary in the example)
```
mfshao, line 177: maybe worth to mention only if the nested data is an array, then the sub-table will be showing

haraprasadj: done, thanks!

PR #388, opened 2020-07-01
chore(qa): Refactor apiUtils

themarcelor:
Addressing all issues pointed out by the linter (eslint).

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
   "devDependencies": {
     "@elastic/elasticsearch": "^7.6.1",
     "@google-cloud/storage": "^4.3.0",
+    "axios": "^0.19.2",
     "atob": "^2.1.2",
     "babel-eslint": "^10.0.3",
     "chai": "^4.2.0",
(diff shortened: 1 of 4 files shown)


suites/portal/indexingPageTest.js
```diff
@@ -61,7 +61,7 @@ Scenario('Navigate to the indexing page and upload a test manifest @indexing', a
   I.waitForElement({ css: '.indexing-page' }, 10);
   I.click('.index-flow-form'); // after clicking open window file upload dialog
   await I.attachFile('input[type=\'file\']', `manifest_${I.cache.UNIQUE_NUM}.tsv`);
-  I.click({ xpath: 'xpath: //button[contains(text(), \'Index Files\')]' });
+  await I.click({ xpath: 'xpath: //button[contains(text(), \'Index Files\')]' });
```
haraprasadj, line 64: I.click does not seem to be asynchronous from the docs ... can you check that once

themarcelor: I've noticed that the `checkPod` log entries were showing up before the click event logs, hence, the introduction of the `await` 🤷

package.json
```diff
@@ -16,6 +16,7 @@
   "devDependencies": {
     "@elastic/elasticsearch": "^7.6.1",
     "@google-cloud/storage": "^4.3.0",
+    "axios": "^0.19.2",
```
haraprasadj, line 19: are we using this in the code? but no harm adding a dependency I guess

PR #423, opened 2020-07-30
chore(qa): Write metrics for data simulator

themarcelor:
tracking data simulator success and failures in influxdb.


mfshao:
i'm not sure, if the script fails in middle of this https://github.com/uc-cdis/gen3-qa/blob/35e327555793e95edfef4cdc2fb3a54933f8af70/jenkins-simulate-data.sh#L86 can it be caught with this metric?

themarcelor:
I have a pushed a change to address your concern @mfshao

PR #439, opened 2020-08-11
fix(qa): Grant RAS consent if expired

themarcelor:
CI tests started failing today because the consent for one of the test users has been revoked.
The test needs to be smart enough to grant the consent and resume the flow when necessary.


itsJiaqi (approved):

haraprasadj (approved):
+1

PR #441, opened 2020-08-14
task(RAS Auth N): Improve aud/scope checks and add refresh token scenario

themarcelor:
Addressing items from TODO list:
```
TODO: Run curl with access token and assert the claim is in place
TODO: Check if ga4gh_passport_v1 shows up inside both the aud & scope blocks of the access token
TODO: Check userinfo output (/user/user) using access token in addition to the browser check
TODO: Try to renew the access token through a refresh token
```

### DIFF
--- a/suites/apis/rasIntegrationTest.js
+++ b/suites/apis/rasIntegrationTest.js
@@ -155,7 +155,6 @@ Scenario('Refresh the access token with the refresh_token obtained through the O
 
   // decode JWT / Access Token
   const accessTokenJson = parseJwt(refreshAccessTokenOutputJson.access_token);
-  expect(accessTokenJson.aud).to.include('ga4gh_passport_v1');
   expect(accessTokenJson.scope).to.include('ga4gh_passport_v1');
 });
 
--- a/suites/apis/rasIntegrationTest.js
+++ b/suites/apis/rasIntegrationTest.js
@@ -121,6 +121,11 @@ Scenario('Use client creds from RAS Test User 1 and auth code to obtain access t
   // decode JWT / Access Token
   const accessTokenJson = parseJwt(tokens.access_token);
 
+  // PXP-6552 - Also decode the id_token and check if the client id is in place
+  const idTokenJson = parseJwt(tokens.id_token);
+  expect(idTokenJson).to.have.property('aud');
+  expect(idTokenJson.aud).to.include(I.cache.rasUser1ClientId);
+
   // run curl with access token and assert the claim is in place
   const userInfoOutputCmd = `curl -H "Authorization: bearer ${tokens.access_token}" https://${process.env.HOSTNAME}/user/user`;
   const userInfoOutput = bash.runCommand(userInfoOutputCmd);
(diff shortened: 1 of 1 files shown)


suites/apis/rasIntegrationTest.js
```diff
@@ -115,11 +115,48 @@ Scenario('Use client creds from RAS Test User 1 and auth code to obtain access t
...
+  const userInfoOutput = bash.runCommand(userInfoOutputCmd);
+  console.log(`userInfoOutput: ${JSON.stringify(userInfoOutput.ga4gh_passport_v1)}`);
+  const userInfoOutputJson = JSON.parse(userInfoOutput);
+
+  expect(userInfoOutputJson).to.have.property('ga4gh_passport_v1');
+  expect(userInfoOutputJson.ga4gh_passport_v1).to.not.to.be.empty;
+
+  console.log(`access token audience: ${accessTokenJson.aud}`);
+  console.log(`access token scope: ${accessTokenJson.scope}`);
+  expect(accessTokenJson.aud).to.include('ga4gh_passport_v1');
```
jiaqi216, line 140: oh this you can get rid of as longer term we don't want scopes in `aud` and is not a requirement from ga4gh

suites/apis/rasIntegrationTest.js
```diff
@@ -115,11 +115,48 @@ Scenario('Use client creds from RAS Test User 1 and auth code to obtain access t
...
+  expect(accessTokenJson.scope).to.include('ga4gh_passport_v1');
```
jiaqi216, line 141: it should have both `ga4gh_passport_v1` and `openid` <- that is if you requested `openid`

suites/apis/rasIntegrationTest.js
```diff
@@ -115,11 +115,48 @@ Scenario('Use client creds from RAS Test User 1 and auth code to obtain access t
...
+
+  const refreshAccessTokenOutputJson = JSON.parse(refreshAccessTokenOutput);
+
+  expect(refreshAccessTokenOutputJson).to.have.property('access_token');
+  expect(refreshAccessTokenOutputJson).to.have.property('id_token');
+  expect(refreshAccessTokenOutputJson).to.have.property('refresh_token');
+
+  // decode JWT / Access Token
+  const accessTokenJson = parseJwt(refreshAccessTokenOutputJson.access_token);
   expect(accessTokenJson.aud).to.include('ga4gh_passport_v1');
```
jiaqi216, line 158: Yeah feel free to remove this now or else we'll just have to remember to remove it should we ever remove scope from aud

suites/apis/rasIntegrationTest.js
```diff
@@ -189,7 +219,16 @@ Scenario('Use client creds for RAS Test User 2 and auth code to obtain access to
...
+  const userInfoOutput = bash.runCommand(userInfoOutputCmd);
+  console.log(`userInfoOutput: ${JSON.stringify(userInfoOutput.ga4gh_passport_v1)}`);
+  const userInfoOutputJson = JSON.parse(userInfoOutput);
+
+  expect(userInfoOutputJson).to.not.have.property('ga4gh_passport_v1');
 
-  console.log(`access token scopes: ${accessTokenJson.aud}`);
+  console.log(`access token audience: ${accessTokenJson.aud}`);
+  console.log(`access token scope: ${accessTokenJson.scope}`);
   expect(accessTokenJson.aud).to.not.include('ga4gh_passport_v1');
```
jiaqi216, line 232: same here

jiaqi216: Oh I guess this is a negative test so would pass either way.

PR #442, opened 2020-08-18
fix(qa): Fix CleverSafe test

themarcelor:
The asynchronous function wasn't reaching the axios GET request and the assertion so this test has been producing false positives 🤦 

### Bug Fixes
- Replace Axios GET request with proper CodeceptJS `I.sendGetRequest` instruction.

**BLOCKED BY**: https://github.com/uc-cdis/cloud-automation/pull/1377

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": null,
     "lines": null
   },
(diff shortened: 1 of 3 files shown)


themarcelor:
Trying a change suggested in:
https://github.com/axios/axios/issues/535

PR #453, opened 2020-09-03
chore : fix portal tests for NCT

atharvar28:
Updating apiUtils to support portal testing for NIAID CLINICAL TRIALS commons

### DIFF
--- a/utils/apiUtil.js
+++ b/utils/apiUtil.js
@@ -429,7 +429,7 @@ module.exports = {
       },
       niaid:{
         summary:{
-          css: '.index_page',
+          css: '.index-page_introduction',
         },
         cards: {
           css: '.g3-button',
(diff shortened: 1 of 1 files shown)


mfshao (approved):

utils/apiUtil.js
```diff
@@ -419,6 +419,22 @@ module.exports = {
...
       },
+      accessclinical: {
+        summary: {
+          css: '.index-page_introduction',
+        },
+        cards: {
+          css: '.g3-button',
+        },
+      },
+      niaid: {
```
paulineribeyre, line 430: we may want to set it to `qa-niaid` to avoid conflicts in other NIAID environments like niaiddata.org?

atharvar28: good suggestion

paulineribeyre (approved):

PR #466, opened 2020-09-22
Chore(qa): set default kubeconfig path

themarcelor:
Small tweak for local test runs to work more seamlessly.


haraprasadj (approved):
Should we run some tests that instrument this, maybe RAS, before merging?

PR #469, opened 2020-09-28
negative tests for ras authn

haraprasadj:


suites/apis/rasAuthN.js
```diff
@@ -0,0 +1,115 @@
...
+ RAS Integration Test - AuthN Only (PXP-6505)
+ This test plan has a few pre-requisites:
+ 1. Fence > 4.23.0 must be deployed.
+ 2. There must be an upstream client previously registered with NIH/RAS and
+    the test environment must have its client & secret ID stored into its Fence config.
+    e.g., $ cat gen3_secrets_folder/configs/fence-config.yaml | yq .OPENID_CONNECT.ras
+          {
+              "client_id": "****",
+              "client_secret": "****",
+              "redirect_url": "{{BASE_URL}}/login/ras/callback"
```
themarcelor, line 12: I guess we should update these comments to make clear that the `discovery_url` is now mandatory. 🤷

suites/apis/rasAuthN.js
```diff
@@ -0,0 +1,115 @@
+/* eslint-disable codeceptjs/no-skipped-tests */
```
themarcelor, line 1: Do we want to have these new scenarios in a whole new/separate file?
`rasAuthN.js` or should we just append these scenarios to the existing `rasIntegrationTest.js` script?

themarcelor (approved):
+1
Left a few comments, but nothing critical.
Thanks 👍

suites/apis/rasAuthN.js
haraprasadj: These are idempotent tests ... we can merge back later by refactoring the existing tests to run independently in the same way. For now, I will leave it here.

PR #478, opened 2020-10-14
Fix classnames of tests to work with new version of portal

mpingram:
### Bug Fixes
- Change classnames in data upload tests to work with new version or portal. (Will continue to work with old versions of portal).


haraprasadj (approved):
+1

PR #501, opened 2020-11-18
chore(qa): Utilize data simulation with imaging_data_file for PR checks that target midrc environments

themarcelor:
There's no `data_file` node in the midrc DD, so we should use `imaging_data_file` instead.

### DIFF
--- a/jenkins-simulate-data.sh
+++ b/jenkins-simulate-data.sh
@@ -25,7 +25,9 @@ function writeMetricWithResult() {
 }
 
 namespace="${1:-${KUBECTL_NAMESPACE:-default}}"
-echo $namespace
+testedEnv="${1:-""}"
+echo "namespace: $namespace"
+echo "testedEnv: $testedEnv"
 export GEN3_HOME="${GEN3_HOME:-${WORKSPACE}/cloud-automation}"
 export TEST_DATA_PATH="${TEST_DATA_PATH:-${WORKSPACE}/testData/}"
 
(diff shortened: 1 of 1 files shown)


atharvar28 (approved):

jenkins-simulate-data.sh
```diff
@@ -62,8 +64,16 @@ fi
 # use "submitted_unaligned_reads" if it's there ...
 #
 leafNode="submitted_unaligned_reads"
-if ! jq -r '.|values|map(select(.category=="data_file"))|map(.id)|join("\n")' < "$TEST_DATA_PATH/schema.json" | grep "$leafNode" > /dev/null; then
-  leafNode="$(jq -r '.|values|map(select(.category=="data_file"))|map(.id)|join("\n")' < "$TEST_DATA_PATH/schema.json" | head -1)"
+
+# diff testedEnvs require diff nodes of type "file"
+if [ "$testedEnv" == "data.midrc.org" ] || [ "$testedEnv" == "qa-midrc.planx-pla.net" ]; then
+  if ! jq -r '.|values|map(select(.category=="imaging_data_file"))|map(.id)|join("\n")' < "$TEST_DATA_PATH/schema.json" | grep "$leafNode" > /dev/null; then
```
paulineribeyre, line 70: why not select the first `_file` node instead?

paulineribeyre: if there's a way to replace `map(select(.category=="data_file"))` with `*file`

themarcelor: We can take this one for a spin and improve it later. 🤷‍♂️ Whatever prevents any force-merges and unblocks midrc work.

PR #505, opened 2020-11-19
fix(midrc): MIDRC doesn't show username at top bar anymore

themarcelor:
quick fix.


mfshao:
`logout` logic should also be update so can bypass midrc env?

mfshao (approved):
lgtm

PR #530, opened 2020-12-15
fix : fixing the homepage for midrc

atharvar28:
As the MIDRC homepage takes more than 10sec to load, hence causing the tests to fail. 

Now the page will refresh itself


themarcelor (approved):
+1

PR #534, opened 2020-12-17
chore:debugging coremetadata test

atharvar28:


themarcelor (approved):
+1 You might want to move the screenshot instruction before the `waitForVisible` call.
Approving anyway if you want to experiment with this.

PR #545, opened 2021-01-05
increase the wait time for the coremetadata page

atharvar28:
increase the wait time for the prop in coremetadata test


themarcelor (approved):
+1

PR #547, opened 2021-01-06
chore(qa): Increase max retries for Data Upload Test

themarcelor:
Trying to avoid intermittent failures caused by slow indexing jobs (potentially due to 1-2 retries).


ocshawn (approved):
looks good to me, but I'm new so may want approval from someone else as well

PR #550, opened 2021-01-13
fix(qa): alternative click approach for dataguids.org GUID resolver

jingh8:
### Bug Fixes
fix the error "element click intercepted"


themarcelor (approved):
+1

PR #566, opened 2021-02-09
chore(qa): Improve logic to capture failures with long-running tests and avoid false negatives

themarcelor:
We have faced a situation recently where a data file upload test failed due to a missing dependency (`metadata-service` from `ssjdispatcher`). The `checkPod` function kept trying to determine whether or not the `indexing` pod had succeeded but it actually presented a `failed` status, which led the function to, frivolously, continue to check the status of the pod and it led to a situation where the Selenium session got reaped by the Selenium Hub and the test produced a misleading result (It failed silently and returned a ZERO `0` exit code).

This anomaly made 2 PRs turn green misleadingly.

This change improves the polling logic to fail-fast and make an exception bubble up to the Jenkins CDIS Lib groovy caller and force-fail the PR check if any similar issues arise in the future.

### DIFF
--- a/hooks/test_results.js
+++ b/hooks/test_results.js
@@ -104,7 +104,7 @@ module.exports = function () {
     if (testResult === undefined) {
       // If there are any Selenium failures, we cannot let the test fail silently.
       // We need to force return an exit code 1.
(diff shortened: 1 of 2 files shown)


emalinowski (approved):

jingh8 (approved):
+1

PR #579, opened 2021-03-12
fix(qa): skip submission page check for midrc

jingh8:
quick fix for midrc

### DIFF
--- a/suites/portal/dataUploadTest.js
+++ b/suites/portal/dataUploadTest.js
@@ -113,22 +113,22 @@ Scenario('Map uploaded files in windmill submission page @dataUpload @portal', a
     users.mainAcct.accessTokenHeader,
   );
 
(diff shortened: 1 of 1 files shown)


mfshao (commented):
any reason why we don't to this in `runTest.sh`?

jingh8:
There is a concern for test coverage for midrc. If we do it in `runTest.sh` , midrc will skip whole dataUpload test suite. But now at least it tests dataUpload through api.

themarcelor (approved):
+1

themarcelor:
Hi @jingh8 @mfshao , exactly. I feel we should try to run, at least, some of the steps for test coverage purposes.
If the absence of that button is the only blocker, we could skip these steps and still test the file upload components (`ssjdispatcher`, `indexs3client`, `fence` Upload PreSigned URL, etc. etc.).

PR #584, opened 2021-03-18
task(ci): Test new mds DELETE endpoint

themarcelor:
Improving test coverage for this new mds feature.


mcannalte (commented):
Looks good to me! Just merged the feature branch so this should be good to go too

mcannalte (approved):

PR #608, opened 2021-04-27
chore(load-testing): Create load testing script for DRS Objects endpoint

themarcelor:
To support the upcoming load testing effort for bdcat.


atharvar28 (approved):

load-testing/indexd/drs-endpoint.js
```diff
@@ -0,0 +1,55 @@
...
+      myFailRate.add(res.status !== 200);
+      if (res.status !== 200) {
+        console.log(`Request response: ${res.status}`);
+        console.log(`Request response: ${res.body}`);
+      }
+      check(res, {
+        'is status 200': (r) => r.status === 200,
+      });
+    });
+    group('wait 0.3s between requests', () => {
```
Avantol13, line 51: what happens if we don't do this? I imagine clients won't add time b/t their requests, they'll just backoff on errors

themarcelor: Sorry for the late reply:
I think the rationale here (at least when we came up with the first exercises) is that we would recommend that customers running automated scripts could set up some pause between requests and also retry on the NGINX RATE LIMIT response (_http 503 - service unavailable - please try again_).

But you're right. We should remove that from the scripts to help us obtain a more accurate result.

PR #620, opened 2021-05-14
set keepSessionAlive to false for etl tests

haraprasadj:
Setting keepSessionAlive to false for ETL tests since no session is created in the test flow


themarcelor (approved):
+1

PR #621, opened 2021-05-17
Chore/automate discovery page

haraprasadj:
Add automated e2e test for discovery page. Initial commit, marked manual for now, feature not yet in master to enable in CI.

### DIFF
--- a/suites/portal/discoveryPageTest.js
+++ b/suites/portal/discoveryPageTest.js
@@ -1,5 +1,4 @@
 const uuid = require('uuid');
-const user = require('../../utils/user.js');
 
 const I = actor();
 I.cache = {};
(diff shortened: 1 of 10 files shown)


suites/portal/discoveryPageTest.js
```diff
@@ -0,0 +1,98 @@
...
+I.cache = {};
+
+Feature('Discovery page');
+
+Scenario('User is able to navigate to Discovery page @manual', ({ discovery }) => {
+  discovery.do.goToPage();
+  discovery.ask.isPageLoaded();
+});
+
+Scenario('Publish a study, search and export to workspace @manual', ({
```
themarcelor, line 13: should we still leave the @manual tag there?
Is that still going to be a semi-automated test?

themarcelor:
@haraprasadj , you should probable leverage a strategy similar to:
```
# studyViewerif [[ $(curl -s "$portalConfigURL" | jq 'contains({studyViewerConfig}) | not') == "true" ]] || [[ ! -z "$testedEnv" ]]; then  donot '@studyViewer'
```
To determine whether or not you want to run the `discoveryPage` annotated test.

themarcelor (approved):
+1 looks good but I'm also leaving a couple of comments.

suites/portal/discoveryPageTest.js
haraprasadj: Like I mentioned above, this is still WIP, so left it as manual for now to prevent it from running accidentally in CI. Once the feature is available in master, I will enable it in CI envs and remove the tag

PR #627, opened 2021-06-03
fixing the revoke access in the after suite

atharvar28:
fixing the revoke access so that it revokes access naturally depending on what kind of env the tests are running on

```
if (process.env.testedEnv.includes('qa-niaid') || process.env.testedEnv.includes('accessclinicaldata')) {
    console.log('### revoking access for user0 in QA/PROD envs ...');
    await bash.runCommand(`
      revoke access
    `);
  } else {
    console.log('### revoking access for user0 in jenkins env ...');
    await bash.runCommand(`
       revoke access
    `);
  }
  console.log('### The access is revoked');
```


themarcelor (approved):
+1

PR #633, opened 2021-06-16
Fix(prod-test): Also scan s3 buckets in access check script

themarcelor:
We should perform the bucket-access-check for both `gs` and `s3` protocols. Not just for Google Buckets.


haraprasadj (approved):
+1

PR #639, opened 2021-06-30
Update README.md

atharvar28:
We do not anymore setup the `selenium` and currently it is done via `docker-compose` file

removing the section so that it doesn't throw off engineer with two setup instructions


jingh8 (approved):

themarcelor (approved):
+1

README.md
```diff
@@ -147,27 +147,6 @@ If there are things which cannot/should not be fetched in the commons by the set
 ```
 In setup, these values will be read and exported to environment variables before running the test. So in the example above `MY_VAR` is actually going to be a string that will need to be parsed when it's used. Note that the setup gives environment variables higher priority over the config file. So if you did `export MY_VAL="rock"` in the shell you're running the test and in your config had `{ MY_VAL: "scissors" }`, when the test is run, `process.env.MY_VAL === "rock"`.
 
-## Local
```
haraprasadj, line 150: maybe these need to be retained

haraprasadj: lines 150 and 151

atharvar28: 👍

haraprasadj (approved):

PR #640, opened 2021-06-30
update home page test for heal bundle

haraprasadj:
Added properties for HEAL bundle where the home page is actually the Discovery page


themarcelor (approved):
+1

PR #641, opened 2021-06-30
Fix : run-test.sh for studyViewerConfig

atharvar28:


themarcelor (approved):
+1

run-tests.sh
```diff
@@ -385,9 +385,9 @@ if [ -z "$checkForPresenceOfMetadataIngestionSowerJob" ]; then
 fi
 
 # studyViewer
-if [[ $(curl -s "$portalConfigURL" | jq 'contains({studyViewerConfig}) | not') == "true" ]] || [[ ! -z "$testedEnv" ]]; then
+if [[ $(curl -s "$portalConfigURL" | jq 'contains({studyViewerConfig}) | not') == "true" ]]; then
```
haraprasadj, line 388: the double quotes around true should be removed, in other similar checks too

haraprasadj (approved):

run-tests.sh
haraprasadj: that did not seem to work, so trying removing `== true` part

haraprasadj (approved):

themarcelor (approved):
+1

PR #644, opened 2021-07-06
PXP-7807 Audit-service tests

paulineribeyre:
Jira Ticket: [PXP-7807](https://ctds-planx.atlassian.net/browse/PXP-7807)

### New Features
- Add tests for the audit-service integration with Fence: download and login events

### Improvements
- Fix bug in indexd tasks "addFileIndices" function causing error messages "file is not defined"
- Refactor RAS client creation logic: remove code duplication, add error logs, and add a random string at the end of the client names to avoid "client already exists" issues
- Refactor "smartWait" function so that it doesn't wait for nothing at the end of the loop


run-tests.sh
```diff
@@ -416,6 +416,16 @@ elif ! (g3kubectl get pods --no-headers -l app=hatchery | grep hatchery) > /dev/
   donot '@exportToWorkspacePortalHatchery'
 fi
 
+#
+# audit-service tests only run for some repos.
+# they don't run for manifest repos because the audit-service might not
+# be deployed everywhere.
+#
+if [[ "$isGen3Release" != "true" && "$service" != "gen3-qa" && "$service" != "fence" && "$service" != "audit-service" && "$service" != "cloud-automation" ]]; then
```
themarcelor, line 424: Don't we want to make sure we have CI with diff mix-and-match of versions in cdis-manifest PRs?
We should be able to figure out whether audit-service is configured against a given environment by checking the manifest versions block and/or the portal/gitops.json.
Just checking, if you prefer this approach, that's ok.

paulineribeyre: sure

atharvar28 (approved):
👍

suites/portal/rasAuthN.js
```diff
@@ -29,6 +29,12 @@ const bash = new Bash();
 const I = actor();
 
 const rasAuthLogin = async (username, password) => {
+  // check these if you run into errors:
+  // - process.env.RAS_TEST_USER_1_USERNAME
+  // - process.env.RAS_TEST_USER_1_PASSWORD
+  expect(username, '"rasAuthLogin" needs "username" to proceed').to.not.be.empty;
+  expect(username, '"rasAuthLogin" needs "password" to proceed').to.not.be.empty;
```
haraprasadj, line 36:
```suggestion
  expect(password, '"rasAuthLogin" needs "password" to proceed').to.not.be.empty;
```

run-tests.sh
```diff
@@ -416,6 +416,29 @@ elif ! (g3kubectl get pods --no-headers -l app=hatchery | grep hatchery) > /dev/
   donot '@exportToWorkspacePortalHatchery'
 fi
 
+#
+# only run audit-service tests for manifest repos IF audit-service is
+# deployed, and for repos with an audit-service integration.
+#
+runAuditTests=true
+if ! [[ "$service" =~ ^(audit-service|fence|cloud-automation|gen3-qa)$ ]]; then
```
haraprasadj, line 424: isn't this check redundant?

haraprasadj (approved):

run-tests.sh
paulineribeyre: No unless i got the logic wrong - if `audit-service|fence|cloud-automation|gen3-qa` then always run the tests, otherwise if `cdis-manifest|gitops-qa|gitops-dev` only run the tests if audit-service is deployed. Don't run the tests for repos that aren't in any of the 2 sets

haraprasadj: got it, sorry misread it, pls ignore

paulineribeyre:
Not ready to merge - I had to revert my Fence config changes in all the Jenkins environments because of "Audit logs are enabled but audit-service is unreachable" errors in cdis-manifest PRs without audit-service. Looking for a solution

themarcelor:
@paulineribeyre , if you want to experiment further in the wild, you can just comment out the `run-tests.sh` logic and leave it as `@donot '@audit'` and label specific PRs in gitops-qa and/pr cdis-manifest with "suites-apis-auditServiceTest" for further experimentation (or even label to select specific jenkins envs or even try that in `ci-env-1` in `qaplanetv2`).

paulineribeyre:
@themarcelor thanks - i think i'm "almost there" 🤞

PR #645, opened 2021-07-08
donot check portal health when not deployed

haraprasadj:
### Improvements
There is now a commons without portal. So, added logic to prevent portal health check when the service is not deployed


jingh8 (approved):
+1

atharvar28 (approved):

PR #646, opened 2021-07-11
Chore/test heal without moon

haraprasadj:
Portal tests failing for heal pr's (consistently for this env, might be flaky for others). The issues noticed are:
1. Timed out receiving message from renderer: 10.000
2. unknown error: cannot activate web view

Implemented some changes based on solutions found online. The tests passed for gitops PR on qa-heal with these changes - https://github.com/uc-cdis/gitops-qa/pull/1490

### DIFF
--- a/services/portal/home/homeTasks.js
+++ b/services/portal/home/homeTasks.js
@@ -15,7 +15,7 @@ module.exports = {
     }
     // wait 2 sec to let the page fully loaded
     I.wait(2);
-    I.captureBrowserLog();
+    // I.captureBrowserLog();
     I.saveScreenshot('Home_page_for_debugging.png');
     portal.seeProp(homeProps.ready_cue, 60);
   },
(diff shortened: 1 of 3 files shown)


services/portal/home/homeTasks.js
```diff
@@ -15,7 +15,7 @@ module.exports = {
     }
     // wait 2 sec to let the page fully loaded
     I.wait(2);
-    I.captureBrowserLog();
+    // I.captureBrowserLog();
```
themarcelor, line 18: This log is useful for debugging. Should we really disable it?

themarcelor (approved):
+1

services/portal/home/homeTasks.js
haraprasadj: This step was failing for HEAL PRs and running locally gave the output something like "No debugger, cannot return variables". I will enable it in a separate PR and do the exercise again with the dummy HEAL PR

PR #649, opened 2021-07-15
fix(guppy-ci): Mutate guppy config to point to Canine indices before running tests

themarcelor:
We need to be able to mutate the guppy config in our CI environments for 2 test flows:

- DYNAMIC data to test the end-to-end flow of the Export to PFB and Export to Workspace features.
- PREDEFINED data to test Guppy queries based on Queries and Expected Responses (based on Canine DD -- ES index pre-hydrated).

This change takes care of the PREDEFINED test flow.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,6 @@
 #!groovy
 
-@Library('cdis-jenkins-lib@master') _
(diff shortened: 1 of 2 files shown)


mfshao (approved):
🎉

PR #652, opened 2021-07-15
getServiceVersion: handle service names with dash

paulineribeyre:
closes #651

### Bug Fixes
- Handle service names with dash in `getServiceVersion`/`runTestsIfServiceVersion`


themarcelor (approved):
+1 I'm not 100% sure the string interpolation will work but we can try that.

themarcelor (approved):
+1

PR #654, opened 2021-07-16
fix bucketname logic

haraprasadj:
### Bug Fixes
We are not fetching the bucket name from config but relying on the standard naming convention. This fixes that.


themarcelor (approved):
+1

PR #657, opened 2021-07-20
Do not run audit-service tests if fence version is too old

paulineribeyre:
### Bug Fixes
- Do not run audit-service tests if fence version is too old


themarcelor (approved):
+1

PR #669, opened 2021-08-09
fix : to evaluate the env for eligibility for running the test

atharvar28:
Change for studyViewer config in `run-tests.sh`  to fix and evaluate when the env is eligible for running the studyViewer test


themarcelor (approved):
+1

PR #690, opened 2021-09-07
fix(ga4gh): use specified GA4GH endpoint, not indexd-specific exposur…

Avantol13:
…e of it

### Improvements
- Use the root/ga4gh/drs/v1/... endpoint (the standard API specified by the GA4GH spec) rather than indexd's exposure via root/index/ga4gh


themarcelor (approved):
+1

PR #697, opened 2021-09-10
fix(ga4gh_drs): no token should result in 401, not 403

Avantol13:
### Improvements
- GA4GH DRS API attempt with no token should result in 401, not 403


suites/apis/drsEndpointsTest.js
```diff
@@ -73,11 +73,18 @@ Scenario('get drs invalid access id @drs', async ({ drs, fence }) => {
...
 
-Scenario('get drs presigned-url no auth header @drs', async ({ drs, fence }) => {
+// New behavior where GA4GH API responds with a 401 with no access token
+Scenario('get drs presigned-url no auth header @drsNew', async ({ drs, fence }) => {
   const signedUrlRes = await drs.do.getDrsSignedUrlWithoutHeader(files.allowed);
   fence.ask.responsesEqual(signedUrlRes, drs.props.noAccessToken);
 });
 
+// Old behavior where GA4GH API responded with a 403 with no access token
+Scenario('get drs presigned-url no auth header @drs', async ({ drs, fence }) => {
```
paulineribeyre, line 83: i don't think this will work as-is because we want to disable the old test when running the new test. Right now if you use `@drsNew` you will run both the old and the new test

services/apis/drs/drsProps.js
```diff
@@ -28,11 +28,13 @@ module.exports = {
...
+  noAccessTokenOld: new Gen3Response({
     request: {},
-    body: { msg: 'Not Authorized. Access Token Required.' },
-    data: { msg: 'Not Authorized. Access Token Required.' },
     status: 403,
   }),
 
+  noAccessToken: new Gen3Response({
+    request: {},
+    status: 401,
```
paulineribeyre, line 38: while running with fence `master`, the `@drsNew` test failed with:
```
expected
{ fenceError: undefined,   parsedFenceError: undefined,   body:
     { msg: 'Not Authorized. Access Token Required.',      status_code: 403 },
   data: 
    { msg: 'Not Authorized. Access Token Required.',
      status_code: 403
 },   status: 403,
   requestMethod: 'GET',
   requestHeaders: undefined,
   requestBody: undefined,
   requestURL: undefined
 }
to have property 'status' of 401, but got 403
```

Avantol13: I think that's b/c this isn't merged https://github.com/uc-cdis/cloud-automation/pull/1726 it's still using the old routing logic

suites/apis/drsEndpointsTest.js
paulineribeyre: #699

PR #699, opened 2021-09-13
fix(ga4gh_drs): no token should result in 401, not 403

paulineribeyre:
### Improvements
- GA4GH DRS API attempt with no token should result in 401, not 403


themarcelor (approved):
+1

paulineribeyre:
closes #697

paulineribeyre:
@themarcelor @Avantol13 The logic is working fine, but the test still fails because it gets 401 instead of 403. It will fail until the cloud-auto PR is merged. It’s kind of an awkward situation, the 2 PRs depend on each other and need to be merged at the same time or the test will fail for other PRs

PR #704, opened 2021-09-23
Feat/drs performance

Avantol13:
Add a test to simulate a client/user attempting to get x number of signed URLs given a list of studies (in the form of `authz` resources).
In other words, we can answer the question "How long does it take to get 10,000 signed URLs across 3 studies in Gen3?"

The test is not designed with parallelized users (there is only 1 virtual user),
as it's meant to simulate a single person trying to get all the data they care about.

**We are assuming the client/user is parallelizing their requests**.
We accomplish that with k6's `batch` support for requests.

**We also assume the client/user is retrying failed requests** (like the rate limiting 503s) with some simple backoff logic
(for each batch of requests, if any failed, keep retrying the failed requests up to 5 times with increasing sleep time in between).

Due to the potential high number of GUIDs that could be requested, we are not able to use the existing
`GUIDS_LIST` that other similar load tests use (as it exceeded the max size of commands in linux - 
it gets passed to a k6 process as a command line arg).
To circumvent this, the logic of paginating and obtaining the GUIDs is built into this test.

There is sufficient flexibility for adjusting num GUIDs, num parallel requests, pagination size, authz resources, etc.

You can see a successful run of this here: https://jenkins.planx-pla.net/job/gen3-run-load-tests/368/console

### New Features
- New GA4GH DRS performance test to align with requirements around upcoming deliverables
- Option to output k6 test results to Datadog


load-testing/ga4gh/drs-performance.js
```diff
@@ -0,0 +1,208 @@
...
+  var params = {
+    'headers': {
+      'Content-Type': 'application/json',
+      'Authorization': `Bearer ${ACCESS_TOKEN}`,
+    },
+  };
+  var body = null;
+
+  if (PASSPORTS_LIST !== undefined && PASSPORTS_LIST !== null && PASSPORTS_LIST !== "") {
+    console.log(`Passport list supplied: ${PASSPORTS_LIST}. Enabling GA4GH Passport flow to POST list of passports rather than use a Gen3 Access Token.`);
```
themarcelor, line 83: Please bear in mind that this function will be triggered by potentially dozens / hundreds of VUs (virtual users) which means that it can be executed hundreds of thousands of times , generating exponentially more log entries because of the `console.log(` call... and that results in a few hundred megabytes of Jenkins disk space that will get filled up.

It should be ok if we don't keep too many jobs & console outputs in the history.
Or you could just comment some of them out 🤷🏼‍♂️ 
not a critical thing. It looks good anyway.

themarcelor (approved):
+1

load-testing/ga4gh/drs-performance.js
Avantol13: yeah, good point. At least for this test there are no plans for hundreds of VUs (or to run this regularly). The test is explicitly meant to simulate a single user (I guess we could do multiple in the future, but I don't expect we'd ever get to hundreds). But I definitely get your point

PR #707, opened 2021-09-27
Fix audit tests: Fence deployment now has a single container

paulineribeyre:
See https://github.com/uc-cdis/cloud-automation/pull/1643

### Bug Fixes
- Fix audit tests: Fence deployment now has a single container

### DIFF
--- a/services/apis/auditService/auditServiceTasks.js
+++ b/services/apis/auditService/auditServiceTasks.js
@@ -66,7 +66,7 @@ async function waitForFenceToRoll() {
       console.log(res);
       let ready = false;
       try {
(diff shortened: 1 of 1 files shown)


themarcelor (approved):
+1

It would be a better approach to, instead of running an assertion based on the health state of all 4 containers, you could just check the status of the overall pod.

How to get the status of the pod:
```
g3kubectl get pod --sort-by=.metadata.creationTimestamp -l app=${labelName} -o jsonpath='{.status.phase}'
```

haraprasadj (approved):

atharvar28 (approved):

PR #722, opened 2021-11-01
fix: wait on aggmds sync job

mfshao:
### Bug Fixes
- AggMDS tests should `await` on `async` job to finish


haraprasadj (approved):
+1

mfshao:
There still occasionally has failues with `received 0 from HEAL` error, but it is much less frequent

Don't have enough time to pin point the cause of that issue, so at least let's first get this fix in

PR #723, opened 2021-11-01
check if pods are terminating

jingh8:
Fix fence pod connection issue in audit service tests

> Error
> Unexpected token ² in JSON at position 0


atharvar28 (approved):

haraprasadj (approved):
+1

PR #725, opened 2021-11-09
check the popup box by the accept button

jingh8:
The popup box of user agreement for MIDRC and VA have different text. To click the Accept button properly, change the code to find if there is a popup box contains the accept button.

### DIFF
--- a/services/portal/home/homeProps.js
+++ b/services/portal/home/homeProps.js
@@ -7,8 +7,11 @@ module.exports = {
   path: '/login',
 
   systemUsePopUp: {
-    locator: {
-      xpath: '//button[contains(text(), \'Accept\')]//ancestor::div[contains(@class, "popup__box")]',
+    locatorForMidrc: {
+      xpath: '//div[contains(text(),"Data Use Agreement for Academic and Institutional Users")]//ancestor::div[contains(@class, "popup__box")]',
+    },
+    locatorForVA: {
+      xpath: '//div[contains(text(),"VA systems are intended for Academic and Institutional Users")]//ancestor::div[contains(@class, "popup__box")]',
     },
   },
 
--- a/services/portal/home/homeProps.js
+++ b/services/portal/home/homeProps.js
@@ -8,7 +8,7 @@ module.exports = {
 
   systemUsePopUp: {
     locator: {
-      xpath: '//div[contains(text(),"Data Use Agreement for Academic and Institutional Users")]//ancestor::div[contains(@class, "popup__box")]',
+      xpath: '//button[contains(text(), \'Accept\')]//ancestor::div[contains(@class, "popup__box")]',
     },
   },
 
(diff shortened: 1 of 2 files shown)


haraprasadj (commented):
The reason for using this was to ensure that the popup that shows up is the Data Use Agreement and not some random popup. Do we want to lose that information?

jingh8:
If we want to keep that, we can add the text of va popup box to the check also.
"VA systems are intended for Academic and Institutional Users"

haraprasadj:
> If we want to keep that, we can add the text of va popup box to the check also. "VA systems are intended for Academic and Institutional Users"

Ok, maybe a more important question is why we cannot have a standard title for data use agreement pop ups across projects 🤷🏼‍♂️

m0nhawk:
> > If we want to keep that, we can add the text of va popup box to the check also. "VA systems are intended for Academic and Institutional Users"
> 
> Ok, maybe a more important question is why we cannot have a standard title for data use agreement pop ups across projects 🤷🏼‍♂️

Each institution will have its own data use agreement, we can't really on it being the same.

Even in QA we should have the environment as close as possible to production.

jingh8:
What about including "Use Agreement" phrase in the title such as this pr: https://github.com/uc-cdis/cdis-manifest/pull/3768/files

mfshao:
For this `systemUseMsg` test we should not hard code them for either VA ori MIDRC or other envs, the test should be generic enough and env independent

We should be able to alter the locater value based on portal config we ingested into the test

m0nhawk:
We should definitely not hardcode it, especially that for most of the projects it's something we need to have "as-is", without any modifications.

haraprasadj:
> > > If we want to keep that, we can add the text of va popup box to the check also. "VA systems are intended for Academic and Institutional Users"
> > 
> > 
> > Ok, maybe a more important question is why we cannot have a standard title for data use agreement pop ups across projects 🤷🏼‍♂️
> 
> Each institution will have its own data use agreement, we can't really on it being the same.
> 
> Even in QA we should have the environment as close as possible to production.

I am not referring to the message in the popup, but the title in the popup itself

m0nhawk:
By "data use agreement" I meant both title & text. We shouldn't change any of it.

haraprasadj:
> For this `systemUseMsg` test we should not hard code them for either VA ori MIDRC or other envs, the test should be generic enough and env independent
> 
> We should be able to alter the locater value based on portal config we ingested into the test

We do have a mechanism to use custom locators like this - https://github.com/uc-cdis/gen3-qa/blob/master/utils/apiUtil.js#L434

The question was more from the portal design perspective, why we cannot standardize the titles of popups like we do for page names

haraprasadj:
> By "data use agreement" I meant both title & text. We shouldn't change any of it.

Hmm ... the locator does not have to be visible text, we should have a way to recognise Data Use Agreement pop up uniquely across implementations, maybe a class name or something else. Otherwise the test code becomes more and more unwieldy and harder to read. I know this is not a current concern, so if we want to deal with it later, it is ok.

haraprasadj (approved):

haraprasadj (approved):
+1

PR #732, opened 2021-11-29
not checking the cards on homePage

atharvar28:
The change here is  ->

commenting out the cards check on commons homePage as not all the commons have cards


haraprasadj (approved):
+1

PR #742, opened 2022-01-12
chore: adding tags to test at Feature level

atharvar28:
Adding Tags to individual tests at Feature level.

What happens here? - so if the service like fence is not deployed to the commons, any test related to fence shouldnt be executed. This tagging will allow that to happen.

### DIFF
--- a/suites/apis/centralizedAuth.js
+++ b/suites/apis/centralizedAuth.js
@@ -1,5 +1,5 @@
 /*eslint-disable */
-Feature('CentralizedAuth');
+Feature('CentralizedAuth @requires-fence');
 /*
 Test sponsor use cases for Gen3 Centralized Authorization with new authorization system:
 
--- a/suites/apis/centralizedAuth.js
+++ b/suites/apis/centralizedAuth.js
@@ -1,5 +1,5 @@
 /*eslint-disable */
-Feature('CentralizedAuth @requires-fence');
+Feature('CentralizedAuth @requires-fence @requires-indexd');
 /*
 Test sponsor use cases for Gen3 Centralized Authorization with new authorization system:
 
--- a/suites/apis/dbgapTest.js
+++ b/suites/apis/dbgapTest.js
@@ -1,5 +1,5 @@
 /*eslint-disable */
-Feature('dbgapSyncing').retry(2);
+Feature('dbgapSyncing @requires-fence').retry(2);
 /*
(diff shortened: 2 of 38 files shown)


suites/apis/metadataIngestionTest.js
```diff
@@ -6,7 +6,7 @@
     within the sower config block (ingest-metadata-manifest & get-dbgap-metadata)
  3. Metadata Service (mds) must also be deployed
 */
-Feature('Metadata Ingestion');
+Feature('Metadata Ingestion @requires-sower');
```
haraprasadj, line 9: does this also require mds / indexd / fence?

suites/apis/exportToWorkspaceTest.js
```diff
@@ -1,4 +1,4 @@
-Feature('ExportToWorkspaceAPITest').retry(2);
+Feature('ExportToWorkspaceAPITest @requires-fence').retry(2);
```
haraprasadj, line 1:
```suggestion
Feature('ExportToWorkspaceAPITest @requires-fence @requires-manifestservice').retry(2);
```

suites/apis/dbgapTest.js
```diff
@@ -1,5 +1,5 @@
 /*eslint-disable */
-Feature('dbgapSyncing').retry(2);
+Feature('dbgapSyncing @requires-fence').retry(2);
```
haraprasadj, line 2:
```suggestion
Feature('dbgapSyncing @requires-fence @requires-indexd').retry(2);
```

suites/apis/centralizedAuth.js
```diff
@@ -1,5 +1,5 @@
 /*eslint-disable */
-Feature('CentralizedAuth');
+Feature('CentralizedAuth @requires-fence');
```
haraprasadj, line 2: does this require arborist too?

suites/apis/metadataIngestionTest.js
atharvar28: This is an initial WIP PR. We can keep adding stuff to this

suites/apis/centralizedAuth.js
atharvar28: It also needs `Indexd`

jingh8 (dismissed):
+1

haraprasadj (approved):

PR #749, opened 2022-01-25
Updating the DD host for test_results

atharvar28:
Updating the DD host from `datadog-agent-cluster-agent.datadog` to `datadog-cluster-agent.datadog` as the datadog deployment was moved to HELM


haraprasadj (approved):
+1

haraprasadj (commented):
We can close this now since we will use DD api instead for CI results

PR #778, opened 2022-03-03
PFB test: fail if the necessary file is not there

paulineribeyre:
fail the test faster and with a more obvious error message

### Improvements
- PFB test: fail if the necessary file is not there


paulineribeyre:
Screen Shot 2022-03-04 at 9 54 45 AM

atharvar28 (approved):

haraprasadj (approved):
+1

paulineribeyre:
force merging because the PFB tests are currently broken

PR #783, opened 2022-03-21
adding RAS DRS tests

atharvar28:
Jira Ticket: [PXP-8795](https://ctds-planx.atlassian.net/browse/PXP-8759)

Adding RAS DRS tests


suites/apis/rasDRSTest.js
```diff
@@ -0,0 +1,122 @@
...
+  const consentURL = I.grabCurrentUrl();
+  I.saveScreenshot('consentURL.png');
+  if (consentURL === `https://${rasServerURL}/auth/oauth/v2/authorize/consent`) {
+    I.click({ xpath: 'xpath: //input[@value=\'Grant\']' });
+  }
+  I.waitForElement({ css: '.auth-list' }, 10);
+  await I.click({ xpath: 'xpath: //button[contains(text(), \'Yes, I authorize.\')]' });
+  await sleepMS(3000);
+  const authCodeURL = await I.grabCurrentUrl();
+  const authCode = authCodeURL.split('=')[1];
```
johnfrancismccann, line 64: Could we parse this param from the url using `code`?
```
const { URL } = require('url');
const querystring = require('querystring');

const url = new URL(authCodeURL);
const authCode = querystring.parse(url.search.substr(1)).code
```

atharvar28: that looks a good idea! it would work when the `code` is anywhere in the URL

suites/apis/rasDRSTest.js
```diff
@@ -0,0 +1,248 @@
...
+  if (deleteRecordReq.status === 200) {
+    console.log('Indexd record is deleted');
+  }
+});
+
+async function getTokens(I) {
+  console.log('Getting the Auth Code ...');
+  I.amOnPage(`https://${rasServerURL}/auth/oauth/v2/authorize?response_type=code&client_id=${I.cache.clientID}&redirect_uri=https://qa-dcp.planx-pla.net/user/login/ras/callback&scope=${scope}&idp=ras`);
+  // GET auth/oauth/v2/authorize to get the rasAuthCode
+  await sleepMS(3000);
```
johnfrancismccann, line 125: Is this to wait for the login form to load? If so, would there be a way to instead wait until either the login form (or entire) DOM loads? That way, if it takes longer than 3 seconds to load, the test will still be able to find the element.

suites/apis/rasDRSTest.js
```diff
@@ -0,0 +1,248 @@
...
+  I.saveScreenshot('NIHLogin_Page.png');
+  // fill in the RAS user creds
+  I.fillField('USER', process.env.RAS_TEST_USER_1_USERNAME);
+  I.fillField('PASSWORD', process.env.RAS_TEST_USER_1_PASSWORD);
+  I.click({ xpath: 'xpath: //button[contains(text(), \'Sign in\')]' });
+  await sleepMS(3000);
```
johnfrancismccann, line 131: Similar to comment above, maybe there is a way to wait until the redirect completes (e.g. `I.amOnPage('https://qa-dcp.planx-pla.net/user/login/ras/callback')`) instead of relying on a specific length of time?

suites/apis/rasDRSTest.js
atharvar28: we can! we can check if any element (eg. Sign-in button) loads instead of waiting for a specific amount of time. I would have do something like `I.waitforElement`.

hooks/runTest_conditions.js
```diff
@@ -0,0 +1,45 @@
...
+        console.log('Skipping the RAS DRS tests since required configuration is not in fence-config.yaml');
+        suite.test.forEach((test) => {
+          test.run = function skip() { // eslint-disable-line func-names
+            console.log(`Ignoring test - ${test.title}`);
+            this.skip();
+          };
+        });
+      }
+
+      const updateCronjob = bash.Command('g3kubectl get cronjob fence-visa-update > /dev/null 2>&1');
```
haraprasadj, line 22:
```suggestion
      const updateCronjob = bash.runCommand('g3kubectl get cronjob fence-visa-update > /dev/null 2>&1');
```

hooks/runTest_conditions.js
```diff
@@ -0,0 +1,45 @@
...
+        console.log('Skipping the test since cronjob is not deployed to the env.');
+        suite.test.forEach((test) => {
+          test.run = function skip() { // eslint-disable-line func-names
+            console.log(`Ignoring test - ${test.title}`);
+            this.skip();
+          };
+        });
+      }
+
+      const cleanupCronjob = bash.Command('g3kubectl get cronjob fence-cleanup-expired-ga4gh-info > /dev/null 2>&1');
```
haraprasadj, line 33:
```suggestion
      const cleanupCronjob = bash.runCommand('g3kubectl get cronjob fence-cleanup-expired-ga4gh-info > /dev/null 2>&1');
```

hooks/runTest_conditions.js
```diff
@@ -0,0 +1,45 @@
...
+        suite.test.forEach((test) => {
+          test.run = function skip() { // eslint-disable-line func-names
+            console.log(`Ignoring test - ${test.title}`);
+            this.skip();
+          };
+        });
+      }
+
+      const updateCronjob = bash.Command('g3kubectl get cronjob fence-visa-update > /dev/null 2>&1');
+      if (!(updateCronjob)) {
```
haraprasadj, line 23: bash.runCommand might return an error message too, so we should check regex here, checking not true may not be enough

atharvar28: 👍

haraprasadj (dismissed):

PR #785, opened 2022-03-28
(PXP-9639): Clean up Elasticsearch indices after ETL and PFB export tests

haraprasadj:
Jira Ticket: [PXP-9639](https://ctds-planx.atlassian.net/browse/PXP-9639)

### ~TODO~ DONE
- [X] merge https://github.com/uc-cdis/gitops-qa/pull/1925

### Improvements
- Clean up ES indices after ETL and PFB Export tests

### DIFF
--- a/services/apis/etl/etlTasks.js
+++ b/services/apis/etl/etlTasks.js
@@ -25,7 +25,6 @@ module.exports = {
         const index = this.getIndexFromAlias(alias);
         console.log(`deleting all index versions associated with alias ${alias}`);
         this.deleteAllIndexVersions(index);
-        expect(this.existAlias(alias), 'Fails to delete alias').to.equal(false);
       }
     });
   },
--- a/services/apis/etl/etlTasks.js
+++ b/services/apis/etl/etlTasks.js
@@ -64,7 +64,7 @@ module.exports = {
   deleteAllIndexVersions(index) {
     const match = index.match(RegExp('(.*)_([0-9]+)$'));
     const prefix = match[1];
-    let number = Number.parseInt(match[2]);
+    let number = Number.parseInt(match[2], 10);
     for (; number >= 0; number -= 1) {
       const indexToDelete = `${prefix}_${number}`;
       if (!this.deleteIndices(`${indexToDelete}`)) {
--- a/services/apis/etl/etlTasks.js
+++ b/services/apis/etl/etlTasks.js
@@ -19,11 +19,11 @@ module.exports = {
     });
 
     aliases.forEach((alias) => {
-      if (etl.do.existAlias(alias)) {
-        const index = etl.do.getIndexFromAlias(alias);
+      if (this.existAlias(alias)) {
+        const index = this.getIndexFromAlias(alias);
         console.log(`deleting all index versions associated with alias ${alias}`);
-        etl.do.deleteAllIndexVersions(index);
-        expect(etl.do.existAlias(alias), 'Fails to delete alias').to.equal(false);
+        this.deleteAllIndexVersions(index);
+        expect(this.existAlias(alias), 'Fails to delete alias').to.equal(false);
       }
     });
   },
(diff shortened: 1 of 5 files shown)


services/apis/etl/etlTasks.js
```diff
@@ -9,6 +9,25 @@ const bash = new Bash();
  * ETL Tasks
  */
 module.exports = {
+  // TODO docstring
+  cleanUpIndices() {
```
haraprasadj, line 13: we should not delete all indices, should leave the ones that are aliased ... otherwise the alias itself might get deleted in the cleanup and prevent guppy from coming up on subsequent runs

suites/apis/etlTest.js
```diff
@@ -1,15 +1,26 @@
...
 
 const bash = new Bash();
+let aliases = [];
 
 BeforeSuite(async ({ etl }) => {
-  etl.props.aliases.forEach(async (alias) => {
-    const index = etl.do.getIndexFromAlias(alias);
-    if (index !== 'error') {
-      etl.do.deleteIndices(index);
+  const etlMappingNames = bash.runCommand(`g3kubectl get cm etl-mapping -o jsonpath='{.data.etlMapping\\.yaml}' | yq '.mappings[].name' | xargs`).split(" ");
```
haraprasadj, line 12: is there any specific reason to do it before and after suite?

suites/apis/etlTest.js
```diff
@@ -31,11 +42,14 @@ Scenario('run ETL second time @etl', async ({ I, sheepdog }) => {
 });
 
 AfterSuite(async ({ etl }) => {
-  etl.props.aliases.forEach((alias) => {
+  console.log("cleaning up indices created by ETL runs");
```
haraprasadj, line 45: can we not just call cleanUp method here?

suites/apis/etlTest.js
johnfrancismccann: `etl.ask.hasVersionIncreased(index, 0);` in `AfterSuite` requires that the index end with `_1`.

It's a bit of an edge case, but if indices exist before `etlTest.js` runs, those indices would be incremented to a version greater than `_1` (e.g. say `jenkins-dcp.planx-pla.net_subject_1` isn't deleted because the pod running the integration tests crashed after the ETL runs but before index cleanup in `AfterSuite`, then on the next `jenkins-dcp` run, `jenkins-dcp.planx-pla.net_subject_2` would be created and `etl.ask.hasVersionIncreased(index, 0);` would fail).

suites/apis/etlTest.js
johnfrancismccann: I really wanted to use `cleanUpIndices` here, but `cleanUpIndices` doesn't ask `hasVersionIncreased`.

services/apis/etl/etlTasks.js
haraprasadj: I meant aliases might get cleaned up by the cloud-auto's garbage collection or AWS' snapshot process if they are not linked to an index, I am not 100% sure how it happened but have had cases where the alias itself goes away and I had to manually recreate them

suites/apis/etlTest.js
haraprasadj: I thought that function's purpose is to check that ETL ran the second time fine, not sure why it is in AfterSuite though, will check the code

johnfrancismccann: I can move `etl.ask.hasVersionIncreased(index, 0);` to the last `Scenario`. That way, we can use `cleanUpIndices` in `AfterSuite`.

services/apis/etl/etlTasks.js
johnfrancismccann: After the gitops-qa PR is merged, it should no longer matter though if those unlinked aliases get cleaned up because that PR specifically makes Guppy not use aliases associated with the names in `etlMapping.yaml`.

johnfrancismccann (approved):

PR #788, opened 2022-04-08
Adding new requestor tests

nss10:
JIRA Ticket : [PXP-8422](https://ctds-planx.atlassian.net/browse/PXP-8422)

##New Features
* Add more integration test cases to requestor service

### DIFF
--- a/services/apis/requestor/requestorTasks.js
+++ b/services/apis/requestor/requestorTasks.js
@@ -21,15 +21,19 @@ module.exports = {
    * @param {string} policyID - policyID of the policy to request/revoke access
    * @param {boolean} revoke - set to true to create a revoke request
    */
-  async createRequestForPolicyID(adminUserTokenHeader, username, policyID, revoke = false) {
-    console.log(`### creating request for a policy id: ${policyID}`);
+  async createRequestForPolicyID(adminUserTokenHeader, username, policyID, revoke = false, policy_status = null) {
+    console.log(`### creating request for a policy id: ${policyID} with revoke set as ${revoke}`);
     const endPoint = revoke ? `${requestorProps.endpoint.requestEndPoint}?revoke` : `${requestorProps.endpoint.requestEndPoint}`;
+    const data = {
+      username,
+      policy_id: policyID,
+    };
+    if(policy_status){
+      data["status"] = policy_status;
+    }
     const getResponse = await I.sendPostRequest(
       endPoint,
-      {
-        username,
-        policy_id: policyID,
-      },
+      data,
       adminUserTokenHeader,
     );
     const responseData = getResponse.data;
--- a/suites/apis/requestorTest.js
+++ b/suites/apis/requestorTest.js
@@ -52,7 +52,7 @@ Scenario('User requests access for a policy followed by sending a Revoke request
   expect(userInfo.data.authz['/requestor_integration_test']).to.deep.to.include({ method: 'access', service: 'jupyterhub' });
 
   // Create a request to 'revoke' a policy that is present in Arborist
-  const revokeResponse = await requestorTasks.createRequestForPolicyID(accessTokenHeader, users.user0.username, 'requestor_integration_test', true);
+  const revokeResponse = await requestorTasks.createRequestForPolicyID(accessTokenHeader, users.user0.username, 'requestor_integration_test', revoke=true);
 
   // Check if the access is revoked -- It shouldn't since it is not signed yet
   userInfo = await fence.do.getUserInfo(user0AccessToken);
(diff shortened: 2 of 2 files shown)


services/apis/requestor/requestorTasks.js
```diff
@@ -21,15 +21,21 @@ module.exports = {
    * @param {string} policyID - policyID of the policy to request/revoke access
    * @param {boolean} revoke - set to true to create a revoke request
    */
-  async createRequestForPolicyID(adminUserTokenHeader, username, policyID, revoke = false) {
-    console.log(`### creating request for a policy id: ${policyID}`);
+  async createRequestForPolicyID(
+    adminUserTokenHeader, username, policyID, revoke = false, policyStatus = null,
```
paulineribeyre, line 25: i think `policyStatus` should be `requestStatus`?

suites/apis/requestorTest.js
```diff
@@ -80,3 +80,29 @@ Scenario('Send a revoke request for a policy the user does not have @requestor',
...
+  users, fence,
+}) => {
+  // Check that the user does not have access to policy 'requestor_integration_test'.
+  let userInfo = await fence.do.getUserInfo(users.user0.accessToken);
+  expect(userInfo.data.authz).to.not.have.property('/requestor_integration_test');
+
+  const { accessTokenHeader } = users.mainAcct;
+  // Create a request for a policy that is present in Arborist
+  const createResponse = await requestorTasks.createRequestForPolicyID(accessTokenHeader, users.user0.username, 'requestor_integration_test', false, 'SIGNED');
+  expect(createResponse).to.have.property('status_code', 201);
```
paulineribeyre, line 94: could you add more comments to explain the special case this is testing, for people who are not familiar with requestor

paulineribeyre (commented):
nice 😄

paulineribeyre (approved):
👍

atharvar28 (dismissed):

paulineribeyre (changes requested):
I just realized this test will fail if Requestor is on an older version. We need to add logic to only run this test for the right versions of Requestor

run-tests.sh
```diff
@@ -252,7 +252,7 @@ runTestsIfServiceVersion "@coreMetadataPage" "portal" "2.20.8"
 runTestsIfServiceVersion "@indexing" "portal" "2.26.0" "2020.05"
 runTestsIfServiceVersion "@cleverSafe" "fence" "4.22.4" "2020.09"
 runTestsIfServiceVersion "@requestor" "requestor" "1.5.0" "2022.02"
-runTestsIfServiceVersion "@requestor-new" "requestor" "1.5.1" "2022.05"
+runTestsIfServiceVersion "@requestorNew" "requestor" "1.5.1" "2022.05"
```
paulineribeyre, line 255: the new feature did not make the 2022.05 cut 😬

nss10: Oh! I can change that, but however I'm having trouble with this design. Since this annotaions will only work at a test level but not the scenario level. I was wondering if I can just change to  `runTestsIfServiceVersion "@requestor" "requestor" "1.5.0" "2022.02"` to `runTestsIfServiceVersion "@requestor" "requestor" "1.5.1"`

paulineribeyre: i didn't get it, isn't a scenario the same as a test?

we can't change it that way because we need to include monthly releases, that's what almost everyone runs now. We could change `requestorNew` to something a bit more explicit, but other than that i think this is the way to do it 🤔 Or the requestor tests will be skipped for releases < 2022.06

paulineribeyre: oh, i see what you mean. i thought the tag system worked at the scenario level, are you sure?
if not, either sounds fine to me, but for #2 it looks like this is not up to date: https://github.com/uc-cdis/requestor/blob/master/pyproject.toml#L3

nss10: I also think, once if we are sure that all the commons moved to `requestor 1.5.1` or `2022.06` we can merge the test suite. Thoughts?

run-tests.sh
```diff
@@ -252,8 +252,8 @@ runTestsIfServiceVersion "@coreMetadataPage" "portal" "2.20.8"
 runTestsIfServiceVersion "@indexing" "portal" "2.26.0" "2020.05"
 runTestsIfServiceVersion "@cleverSafe" "fence" "4.22.4" "2020.09"
 runTestsIfServiceVersion "@requestor" "requestor" "1.5.0" "2022.02"
-runTestsIfServiceVersion "@requestorNew" "requestor" "1.5.1" "2022.05"
 runTestsIfServiceVersion "@requestor" "arborist" "3.2.0" "2021.12"
+runTestsIfServiceVersion "@requestorNew" "requestor" "1.5.1" "2022.06"
```
paulineribeyre, line 256: we might need to duplicate the arborist check too :/

suites/apis/requestorNewTest.js
```diff
@@ -0,0 +1,48 @@
+Feature('RequestorNewAPI @requires-requestorNew');
+
+/**
+  This set of integration tests are the extension to the existing
+  requetorTest suite. They test the flow of new features included
```
paulineribeyre, line 5:
```suggestion
  requestorTest suite. They test the flow of new features included
```

run-tests.sh
nss10: That makes sense!

paulineribeyre (approved):

PR #796, opened 2022-04-20
adding executableTests for ras cronjob validation

atharvar28:
Adding executable tests for validating `fence-visa-update` cronjob

### DIFF
--- a/suites/apis/rasCronjobValidate.js
+++ b/suites/apis/rasCronjobValidate.js
@@ -13,11 +13,13 @@ Scenario('Validate fence-visa-update job @manual', ifInteractive(
             3. Check ga4gh_visa_v1 table by command 'gen3 psql fence'
                 - number of visas for the user
                 - check the JTI and expiration on the visa
+            (SQL QUERY - 'SELECT user_id, ga4gh_visa, expires FROM ga4gh_visa_v1 WHERE user_id=<RAS user id> ORDER BY expires DESC;' )
             4. Run 'gen3 job run fence-visa-update'
             5. Run 'gen3 psql fence'
             6. Again check ga4gh_visa_v1 table for the new entry
                 - number of visa(s) for the user (the number should be more)
-                - the JTI(s) and expiration(s) on the new visa(s) should be different 
+                - the JTI(s) and expiration(s) on the new visa(s) should be different
+            (SQL QUERY - 'SELECT user_id, ga4gh_visa, expires FROM ga4gh_visa_v1 WHERE user_id=<RAS user id> ORDER BY expires DESC;' ) 
         `);
     expect(result.didPass, result.details).to.be.true;
   },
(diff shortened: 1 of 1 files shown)


suites/apis/rasCronjobValidate.js
```diff
@@ -0,0 +1,19 @@
...
+const { interactive, ifInteractive } = require('../../utils/interactive.js');
+
+const TARGET_ENVIRONMENT = process.env.GEN3_COMMONS_HOSTNAME;
+
+Scenario('Validate fence-visa-update job @manual', ifInteractive(
+  async () => {
+    const result = await interactive(`
+            1. Get a new passport from /userinfo endpoint
+            2. Log into ${TARGET_ENVIRONMENT} with RAS Login (RAs user and password)
+            3. run `gen3 job run fence-visa-update`
```
Avantol13, line 13: should check visa table before job is run, note number of visas for your user - perhaps note the JTI and expiration

suites/apis/rasCronjobValidate.js
```diff
@@ -0,0 +1,19 @@
...
+            4. run `gen3 psql fence`
+            5. check ga4gh_visa_v1 table for the new entry
```
Avantol13, line 15: now double check the number of visas for your user again, you should have more. and the new one(s) should have different JTIs and expirations than from when you checked before

suites/apis/rasCronjobValidate.js
atharvar28: 👍  let me add that steps and also some description on what needs to be done here while checking the db

suites/apis/rasCronjobValidate.js
atharvar28: adding step to check the visa table after the job is run for different JTI and expiration on the new visa

suites/apis/rasCronjobValidate.js
```diff
@@ -0,0 +1,24 @@
...
+const TARGET_ENVIRONMENT = process.env.GEN3_COMMONS_HOSTNAME;
+
+Scenario('Validate fence-visa-update job @manual', ifInteractive(
+  async () => {
+    const result = await interactive(`
+            1. Get a new passport from RAS /userinfo endpoint
+            2. Log into ${TARGET_ENVIRONMENT} with RAS Login (RAS user and password)
+            3. Check ga4gh_visa_v1 table by command 'gen3 psql fence'
+                - number of visas for the user
+                - check the JTI and expiration on the visa
```
johnfrancismccann, line 15: Maybe we should specify that the JTI/expiration check should be on the latest visa. Also, I don't think it would hurt to document a SQL query for this step (e.g. `SELECT ga4gh_visa, expires FROM ga4gh_visa_v1 ORDER BY expires DESC LIMIT 1;`)

atharvar28: i have added the SQL queries on the both step 3 and step 6 so that it helps the user

johnfrancismccann (approved):

PR #837, opened 2022-08-04
add afterSuite to delete requests

atharvar28:
Adding `AfterSuite` to delete the request created by the test after the test is executed

### DIFF
--- a/suites/apis/requestorNewTest.js
+++ b/suites/apis/requestorNewTest.js
@@ -27,7 +27,9 @@ Before(async ({
 
 AfterSuite(async ({ I }) => {
   console.log('Deleting all the requests created by this test with user cdis.autotest@gmail.com...');
-  console.log(I.cache.requestDidList);
+  if (process.env.DEBUG === 'true') {
+    console.log(I.cache.requestDidList);
+  }
   I.cache.requestDidList.forEach(async (request) => {
     const deleteRequest = await requestorTasks.deleteRequest(request);
     if (deleteRequest === undefined) {
--- a/suites/apis/requestorNewTest.js
+++ b/suites/apis/requestorNewTest.js
@@ -11,6 +11,13 @@ Feature('RequestorNewAPI @requires-requestorNew');
 const { expect } = require('chai');
 const requestorTasks = require('../../services/apis/requestor/requestorTasks.js');
 
+BeforeSuite(async ({
+  I,
+}) => {
+  I.cache = {};
+  I.cache.requestDidList = [];
+});
+
 Before(async ({
   home, users, login,
 }) => {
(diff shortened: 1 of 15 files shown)


suites/apis/requestorNewTest.js
```diff
@@ -11,15 +11,33 @@ Feature('RequestorNewAPI @requires-requestorNew');
...
 Before(async ({
   home, users, login,
 }) => {
   home.do.goToHomepage();
   login.complete.login(users.mainAcct);
 });
 
+AfterSuite(async ({ I }) => {
+  console.log('Deleting all the requests created by this test with user cdis.autotest@gmail.com...');
+  console.log(I.cache.requestDidList);
```
haraprasadj, line 30: can we add a check to print these values only in debug mode

suites/apis/requestorTest.js
```diff
@@ -10,13 +10,31 @@ Feature('RequestorAPI @requires-requestor');
...
 Before(async ({
   home, users, login,
 }) => {
   home.do.goToHomepage();
   login.complete.login(users.mainAcct);
 });
 
+AfterSuite(async ({ I }) => {
+  console.log('Deleting all the requests created by this test with user cdis.autotest@gmail.com...');
+  console.log(I.cache.requestDidList);
```
haraprasadj, line 29: here too

suites/apis/requestorNewTest.js
atharvar28: oh yes, I can do that

haraprasadj (dismissed):
+1

suites/apis/requestorNewTest.js
```diff
@@ -11,15 +11,35 @@ Feature('RequestorNewAPI @requires-requestorNew');
...
 });
 
+AfterSuite(async ({ I }) => {
+  console.log('Deleting all the requests created by this test with user cdis.autotest@gmail.com...');
+  if (process.env.DEBUG === 'true') {
+    console.log(I.cache.requestDidList);
+  }
+  I.cache.requestDidList.forEach(async (request) => {
+    const deleteRequest = await requestorTasks.deleteRequest(request);
+    if (deleteRequest === undefined) {
```
nss10, line 35: Just curious, what would be the result of the test, when the test fails?

suites/apis/requestorNewTest.js
```diff
@@ -40,7 +61,8 @@ Scenario('User requests access for a policy with a signed status (and revoke it
 
   // Create a request to 'revoke' a policy that is present in Arborist again
   // with request status set to SIGNED
-  await requestorTasks.createRequestForPolicyID(accessTokenHeader, users.user0.username, 'requestor_integration_test', true, 'SIGNED');
+  const createResponseSecond = await requestorTasks.createRequestForPolicyID(accessTokenHeader, users.user0.username, 'requestor_integration_test', true, 'SIGNED');
```
nss10, line 64: I feel like I'm probably nit-picking, but maybe you can change the name of this variable to `createRevokeResponse`?

nss10 (changes requested):

suites/apis/requestorNewTest.js
atharvar28: so I have added a return to `deleteRequest` function. So now we look at the status of the deletion and say the request id is deleted.

haraprasadj: I think this only checks that request was triggered which will be true mostly, we should check the status code of the response or something

nss10 (approved):
Looks Good To Me.

PR #840, opened 2022-08-05
Disable dicom viewer test and register user test temporarily

jingh8:


tyro-ops (approved):
LGTM , Let's debug that why these test are failing.

PR #844, opened 2022-08-12
GWAS Test Fixes

tyro-ops:
Pushing fixes to GWAS_frontend_tests feature branch


services/portal/GWAS/GWASTasks.js
```diff
@@ -76,8 +76,9 @@ module.exports = {
   },
 
   async selectDifferentType() {
-    I.seeElement(GWASProps.selectDifferentType);
-    I.click(GWASProps.selectDifferentType);
+    I.seeElement(GWASProps.SelectDifferentType);
+    I.click(GWASProps.SelectDifferentType);
+    I.wait(2);
     I.seeElement(GWASProps.AreyousurePopup);
```
haraprasadj, line 82: can we change this to waitForElement and add a wait time of 2 there instead of an explicit wait?

suites/portal/GWASUITest.js
```diff
@@ -29,9 +29,12 @@ Scenario('GWAS submit workflow through Case Control GWAS @GWASUI', async ({
   await GWASTasks.clickNextButton();
   // review
   await GWASTasks.clickNextButton();
+  // Add Custom
+  await GWASTasks.clickNextButton();
   // TODO: Check the covariates are correct
   // TODO: handle error when select hare group
   // select parameters (group)
+  I.wait(2);
```
haraprasadj, line 37: do we need this explicit wait? can we instead look for some element dyncamically?

services/portal/GWAS/GWASTasks.js
tyro-ops: Actually have tried with waitForElement but it not worked, only then have used hard wait.

suites/portal/GWASUITest.js
tyro-ops: have removed this explicit wait, As it part of debugging.

haraprasadj (approved):

PR #848, opened 2022-08-15
fixing nonDBGaP project usersync tests

atharvar28:


services/apis/nondbgapUsersync/nonDbgapTasks.js
```diff
@@ -106,36 +108,29 @@ module.exports = {
...
-
+  async presignedURLRequestS3(fence, project, projectFile, user, token) {
     let projectFileContents3 = null;
-    let projectFileContentgs = null;
-
-    try {
+    console.log(`Creating presigned url with ${user} user for ${project} after running usersync`);
+    const signedURLs3 = await fence.do.createSignedUrl(projectFile, ['protocol=s3'], token);
+    if (signedURLs3.status !== 200) {
+      console.log(`User ${user} with access can not create s3 signed urls and read the file for ${project}`);
```
BinamB, line 116: I think it would be helpful to output the request log too. Sometimes, it might not be that the user doesn't have access to it. It might be that something went wrong with indexd record creation and we're getting a 404.

services/apis/nondbgapUsersync/nonDbgapTasks.js
```diff
@@ -106,36 +108,29 @@ module.exports = {
...
     }
+  },
 
-    try {
+  async presignedURLRequestGs(fence, project, projectFile, user, token) {
+    let projectFileContentgs = null;
+    console.log(`Creating presigned url with ${user} user for ${project} after running usersync`);
+    const signedURLgs = await fence.do.createSignedUrl(projectFile, ['protocol=gs'], token);
+    if (signedURLgs.status !== 200) {
+      console.log(`User ${user} with access can not create gs signed urls and read the file for ${project}`);
```
BinamB, line 129: Same as previous comment. Might not always be a 401 error so we should output the request message so its easier to debug

suites/apis/nonDbgapUsersyncTest.js
```diff
@@ -139,83 +137,91 @@ AfterSuite(async ({ indexd }) => {
...
-  const dbGapProjectDID = await nondbgap.do.checkDbGapAccess(rasAccessToken);
-  const dbGpProjectAccess = await fence.do.createSignedUrl(dbGapProjectDID, rasAccessToken);
-  if (dbGpProjectAccess.data.status === 200) {
-    console.log(`The presigned url for project 12345 files is created. The URL -> ${dbGpProjectAccess.data.urls}`);
-  }
-
-  // running usersync job after the test
-  await nondbgap.do.runUserSync();
-});
+// Scenario('Presigned Url with RAS user @nondbgapUsersyncTest', async ({ fence, users, nondbgap }) => {
```
BinamB, line 186: Do we want to uncomment these? Have we tested with the RAS user?

services/apis/nondbgapUsersync/nonDbgapTasks.js
atharvar28: I agree, having `request.status` on failure can help debugging.

BinamB (approved):

PR #858, opened 2022-10-07
fix: accept system use agreement for VA envs (va or mickey)

m0nhawk:
Jira Ticket: N/A

### Bug Fixes
* fix for system use agreement for VA and qa-mickey envs

### DIFF
--- a/services/portal/login/loginTasks.js
+++ b/services/portal/login/loginTasks.js
@@ -1,6 +1,8 @@
 const loginProps = require('./loginProps.js');
 const portal = require('../../../utils/portal.js');
+const { Bash } = require('../../../utils/bash.js');
 
+const bash = new Bash();
 const I = actor();
 
 /**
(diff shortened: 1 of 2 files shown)


services/portal/login/loginTasks.js
```diff
@@ -21,6 +34,9 @@ module.exports = {
     if (I.seeCurrentUrlEquals(loginProps.path)) {
       console.log('Already on Login Page');
     } else {
+      if (process.env.testedEnv.includes('mickey') || process.env.testedEnv.includes('va')) {
+        await this.systemUseMsg();
```
atharvar28, line 38: you can just do a `await home.do.systemUseMsg();` here within the if loop , so that we dont have duplicate code for systemUseMsg

services/portal/login/loginTasks.js
```diff
@@ -12,6 +12,19 @@ module.exports = {
...
+    const title = await bash.runCommand('gen3 secrets decode portal-config gitops.json | jq \'.components.systemUse.systemUseTitle\'');
+    console.log(title);
+    if (title !== null && title !== '') {
+      const numberOfElements = await I.grabNumberOfVisibleElements(`//div[contains(text(), ${title})]//ancestor::div[contains(@class, "popup__box")]`);
+      console.log(`### numberOfElements:${numberOfElements}`);
+      if (numberOfElements > 0) {
+        I.click(homeProps.systemUseAcceptButton.locator);
+      }
+    }
+  },
```
atharvar28, line 26: if you do use the suggestion from below, you can remove this function

PR #865, opened 2022-11-07
fix(install-pypfb): ensure pip is up to date

Avantol13:
Nightly build test PFB Export `@requires-portal @e2e Install the latest pypfb CLI version and make sure we can parse the avro file @pfbExport – Install the latest pypfb CLI version and make sure we can parse the avro file @pfbExport` failed twice in a row when using the BDC environment last night and today.

Looks like the cryptography/Rust compiler issue. The test is using Python 3.8 (instead of 3.9, not sure why) but the simple solution is to ensure that we upgrade pip before attempting to install pypfb in the test. I have a PR to do so here: 
```
  Failed to build cryptography
  ERROR: Could not build wheels for cryptography which use PEP 517 and cannot be installed directly
  WARNING: You are using pip version 19.2.3, however version 22.3.1 is available.
  You should consider upgrading via the 'pip install --upgrade pip' command.
```

### Bug Fixes
- Fix broken nightly build by ensuring pip is updated before installing pypfb


haraprasadj (approved):
+1

PR #870, opened 2022-11-10
separate utils function for pre-release tests

haraprasadj:
Jira Ticket: [QAT-598](https://ctds-planx.atlassian.net/browse/QAT-598)


utils/oidcUtils.js
```diff
@@ -0,0 +1,54 @@
...
+      return 'Could not find nonce. Make sure your id_token is not expired.';
+    }
+    return parseInt(nonceStr.split('-')[2], 10);
+  } catch (e) {
+    console.error(e);
+    return null;
+  }
+}
+
+async function runVerifyNonceScenario(nonceVal) {
```
atharvar28, line 30: you are adding a manual interactive test in utils. Its more test specific than common util function

tyro-ops: No , getting out

PR #873, opened 2022-11-15
fixing the Auth tests in jenkins

atharvar28:


paulineribeyre (approved):
We can merge this first, but in the light of this: https://cdis.slack.com/archives/C0294A7KVE0/p1668526784117479?thread_ts=1668440012.662969&cid=C0294A7KVE0
i think we should also add a check, and if the call was not successful, print the error logs instead of trying to parse it into `(client id, client secret)`

paulineribeyre (approved):

PR #875, opened 2022-11-21
remove lint-diff and update eslintdiff script

haraprasadj:


atharvar28 (approved):

.travis.yml
```diff
@@ -2,7 +2,7 @@ language: node_js
 git:
   depth: false
 node_js:
-  - "10"
```
mfshao, line 5: why not 18?

.travis.yml
```diff
@@ -2,7 +2,7 @@ language: node_js
...
+  - 17
```
mfshao, line 5: why not 18?

haraprasadj: Travis fails with 18

haraprasadj: https://app.travis-ci.com/github/uc-cdis/gen3-qa/builds/258034794

mfshao: I think we could upgrade the travis to use ubuntu 18 as base image and use nodejs 18, so travis and docker are in sync

if that doesn't work I'd suggest to downgrage to node 16, because only even-numbered nodejs versions are LTS versions

PR #876, opened 2022-11-22
fix `after all` hooks

haraprasadj:
- Use node 14
- Update libraries
- Update package and package-lock
- Add exception handling to all Before and After blocks where there are no explicit checks


suites/apis/dbgapTest.js
```diff
@@ -100,45 +100,49 @@ const new_dbgap_records = {
 };
 
 BeforeSuite(async ({ fence, users, indexd }) => {
```
paulineribeyre, line 102: missing a `try/except` here maybe?

run-tests.sh
```diff
@@ -591,8 +591,8 @@ if [[ "$(hostname)" == *"cdis-github-org"* ]] || [[ "$(hostname)" == *"planx-ci-
   # Start selenium process within the ephemeral jenkins pod.
   # npx selenium-standalone install --version=4.0.0-alpha-7 --drivers.chrome.version=96.0.4664.45 --drivers.chrome.baseURL=https://chromedriver.storage.googleapis.com
   # timeout $seleniumTimeout npx selenium-standalone start --version=4.0.0-alpha-7 --drivers.chrome.version=96.0.4664.45 &> selenium.log &
-  npm install selenium-standalone --save-dev
-  npx selenium-standalone install && npx selenium-standalone start &
+  # npm install selenium-standalone --save-dev
+  timeout $seleniumTimeout npx selenium-standalone install && npx selenium-standalone start &
```
paulineribeyre, line 595: could you explain why we need the changes in these 2 lines?

emalinowski (approved):

PR #926, opened 2023-03-17
StudyRegistration Tests

atharvar28:
[GQE-26](https://ctds-planx.atlassian.net/browse/GQE-26)

[GQE-26]: https://ctds-planx.atlassian.net/browse/GQE-26?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mfshao (commented):
very cool

I have a generic question: The request access for study registration flow will create a kayako ticket, therefore it requires an operational kayako wrapper service. How would we want to address this in our CI?

smvgarcia:
The kayako ticket serves 2 purposes during study registration: (1) it communicates the information necessary for User Services to approve the access request; and (2) it alerts User Services that someone has requested access, so we can reach out to provide further resources and instruction.

I think what Atharva did was to automate #1 such that his script retrieves the information that normally is sent in the Kayako ticket to automatically approve access to register. Because this is just for testing, we do not need to do #2. Therefore, his automation means that -- for testing purposes only -- there does not need to be a kayako ticket created. If we then go on to use this code for automating real study registration access requests, however, we will need to find a way to make sure User Services knows about the request so we can proceed with #2.

suites/portal/studyRegistrationTest.js
```diff
@@ -0,0 +1,107 @@
+Feature('Study Registration @requires-portal @requires-requestor @aggMDS @discoveryPage');
```
mfshao, line 1: I think this test should be only run for HEAL manifests, or more specifically, for any manifest that has `"studyRegistration": true,` in feature flags. Otherwise this test will be failing

mfshao (approved):

haraprasadj (dismissed):

haraprasadj (approved):

PR #930, opened 2023-03-27
Do not look for core metadata citation on portal page

paulineribeyre:
Jira Ticket: [PXP-xxxx](https://ctds-planx.atlassian.net/browse/PXP-xxxx)

### Improvements
- Do not look for core metadata citation on portal page. This is a temporary fix because the portal does not always display the new citation format, depending on the deployed version of portal/peregrine/pidgin


haraprasadj (approved):
+1

PR #947, opened 2023-05-01
fix(google): fix linking test false failures because of relative URL

Avantol13:
### Bug Fixes
- fix linking test false failures because of relative URL


haraprasadj (dismissed):
+100

PR #962, opened 2023-05-19
Improve core metadata test errors

paulineribeyre:
### Improvements
- Improve core metadata test errors: check that the request succeeded before checking the fields in the response


haraprasadj (approved):
+1

PR #967, opened 2023-06-06
Correctly check array values in file page

paulineribeyre:
see https://github.com/uc-cdis/data-portal/pull/1327

### Improvements
- Correctly check array values in file page


haraprasadj (approved):
+1

PR #972, opened 2023-06-13
upgrading dependencies to use new version of ansi-regex

atharvar28:
Jira Ticket: [PXP-307](https://ctds-planx.atlassian.net/browse/PPS-307)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,13235 @@
 {
   "name": "gen3-qa",
   "version": "0.1.0",
-  "lockfileVersion": 1,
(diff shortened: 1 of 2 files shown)


paulineribeyre (commented):
i still see more old versions in the lock file, here are the first ones:
```
gauge.ansi-regex 2.1.1
gauge.strip-ansi.ansi-regex ^2.0.0
has-ansi.ansi-regex 2.1.1
localtunnel.ansi-regex 2.1.1
string-length.ansi-regex 2.1.1
```

PR #982, opened 2023-07-21
Adding gen3-client tests

atharvar28:
Jira Ticket: [GQE-102](https://ctds-planx.atlassian.net/browse/GQE-102)

### Deployment changes

[GQE-102]: https://ctds-planx.atlassian.net/browse/GQE-102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


haraprasadj (approved):

suites/apis/dataclientTest.js
```diff
@@ -0,0 +1,253 @@
...
+const fs = require('fs');
+const { execSync } = require('child_process');
+const axios = require('axios');
+const AdmZip = require('adm-zip');
+const users = require('../../utils/user');
+// const os = require('os');
+
+const I = actor();
+
+const latestGen3client = 'https://github.com/uc-cdis/cdis-data-client/releases/latest/download/dataclient_linux.zip';
```
mfshao, line 27: not sure I get the idea of this CI: looks like it is download the binary from the latest release of gen3-client instead of from the current branch/tag? So it mean even with this tests in CI, we cannot catch regressions in gen3-client before commiting it into master and make a release?

suites/gen3client/gen3clientTest.js
```diff
@@ -0,0 +1,174 @@
...
+  const version = execSync(`${I.cache.goPath}/bin/gen3-client --version`, { shell: 'bash' });
+  output.debug(`### gen3-client version installed is ${version}`);
+});
+
+AfterSuite(async ({ I, files, indexd }) => {
+  // delete the fileUploaded and creds.file
+  output.log('Deleting test upload files ...');
+  files.deleteFile(I.cache.fileToBeUploaded);
+  output.log('Deleting creds.file ...');
+  files.deleteFile(I.cache.credsFile);
```
mfshao, line 73: log/profile/stats files for gen3-client exists in `~/.gen3`, you may want to clean it up as well, but also ok to leave as-is I think

haraprasadj: everything runs in the PR pod (not in the adminvm) which gets removed after the run, so cleanup is only needed if other tests get affected which isn't the case here I think.

mfshao (dismissed):

haraprasadj (approved):

PR #999, opened 2023-09-29
removing the "doc_type" field 

EliseCastle23:
removing the "doc_type" field from the index mapping as ES7 no longer supports it and we are moving Jenkins over to ES7.


haraprasadj (approved):
+1

PR #1007, opened 2023-10-27
Remove data-simulator python 3.8 line

paulineribeyre:
Data simulator now uses python 3.9, see https://github.com/uc-cdis/data-simulator/pull/99


haraprasadj (approved):
+1

PR #1023, opened 2023-12-14
Print logs of failed pods after giving up

paulineribeyre:
### Improvements
- Print logs of failed pods after giving up


haraprasadj (approved):
+1

PR #1053, opened 2024-08-23
BIH-39 Fix submit tests for BIH

paulineribeyre:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/BIH-39

Fixing the tests for an edge case found in the BIH dictionary.

Screenshot 2024-08-23 at 1 40 55 PM

The tests run this data-simulator command, using `image_annotation` because that's the "leaf node" the bash script found:
> Running command: poetry run data-simulator submission_order --url https://s3.amazonaws.com/dictionary-artifacts/bih_dictionary/0.1.0/schema.json --path /home/jenkins/agent/workspace/DIS_GitHub_Org_gitops-qa_PR-2968/testData/ --node_name image_annotation

The output is:
```
  DataImportOrderPath.txt:
project	administrative
dataset	administrative
subject	administrative
imaging_study	medical_imaging
imaging_series	medical_imaging
core_metadata_collection	administrative
image_file	data_file
image_annotation	data_file
```

The `getPathWithFileNode` function looks for a file node in this output ^. It finds `image_file` first. The rest of the submission logic, specifically [this](https://github.com/uc-cdis/gen3-qa/blob/3d16fe445bebeb433ce5b1313fd5785feadeda55/utils/nodes.js#L152), assumes that the file node is a "leaf node" (a node with no child links), which is not the case here (`image_file` has a child link to `image_annotation`). This creates problems later (`image_file` is [removed](https://github.com/uc-cdis/gen3-qa/blob/3d16fe445bebeb433ce5b1313fd5785feadeda55/utils/nodes.js#L152) from the path since it's assumed to be a leaf node, so it's not submitted before `image_annotation`, so the submission fails).

This can be fixed by picking the _last_ file node instead of the _first_ file node in `getPathWithFileNode`. The last file node of the list should be a leaf node.

### Improvements
- Remove the unused `Node.target` property
- Fix `getPathWithFileNode` to return the last file node it finds, instead of the first one, since the rest of the submission logic assumes it's a leaf node

### DIFF
--- a/services/apis/sheepdog/sheepdogTasks.js
+++ b/services/apis/sheepdog/sheepdogTasks.js
@@ -150,10 +150,7 @@ module.exports = {
    */
   async addNodes(nodesList, accessTokenHeader = user.mainAcct.accessTokenHeader) {
     // add nodes, in sorted key ascending order
-    console.log('sheepdogTasks.addNodes nodesList BEFORE sort: ', nodesList);
-    console.log('sheepdogTasks.addNodes nodesList AFTER sort:', nodes.sortNodes(nodesList));
     for (const node of nodes.sortNodes(nodesList)) {
-      console.log('sheepdogTasks.addNodes submitting node: ', node.data.type);
       await this.addNode(node, accessTokenHeader);
     }
   },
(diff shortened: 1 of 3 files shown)


paulineribeyre:
This should address problems in https://github.com/uc-cdis/gitops-qa/pull/2968 and https://github.com/uc-cdis/cdis-manifest/pull/7618

haraprasadj (approved):

PR #1065, opened 2025-02-03
Handle Retrieval of Pod Name

krishnaa05:
Link to JIRA ticket if there is one:


paulineribeyre (changes requested):
See https://cdis.slack.com/archives/C04LAGYRQ59/p1738623868656979?thread_ts=1738615530.227379&cid=C04LAGYRQ59

